PR #20, opened 2020-07-14
Add apis unittest

yinchimaoliang:
Add unittests.


mmdet3d/apis/inference.py
```diff
@@ -76,8 +76,8 @@ def inference_detector(model, pcd):
         # scatter to specified GPU
         data = scatter(data, [device.index])[0]
     else:
-        raise NotImplementedError('Not support cpu-only currently')
-
+        data['img_metas'] = data['img_metas'][0].data
```
ZwwWayne, line 79: Add comments here, indicating that this is a workaround to avoid the bug of `MMDataParallel`.

PR #41, opened 2020-07-25
Add ply doc

yinchimaoliang:
Add doc to convert file from ply to bin.


docs/getting_started.md
```diff
@@ -213,7 +213,29 @@ Examples:
...
+    property_names = data[0].dtype.names  # read names of properties
+    for i, name in enumerate(
+            property_names):  # read data by property
+        data_np[:, i] = data_pd[name]
+    data_np.astype(np.float32).tofile(output_path)
+```
+Examples:
 
+```shell
+convert('./test.ply', './test.bin')
```
ZwwWayne, line 237: If it is a shell script, it should be `python convert.py`. If you only want to use this function, it should be `convert_ply(xxxx)`. We also need to remind the users that this function can also be used for data preprocessing for training `ply data`.

docs/getting_started.md
```diff
@@ -213,7 +213,29 @@ Examples:
...
+    data_np = np.zeros(data_pd.shape, dtype=np.float)  # initialize array to store data
+    property_names = data[0].dtype.names  # read names of properties
+    for i, name in enumerate(
+            property_names):  # read data by property
+        data_np[:, i] = data_pd[name]
+    data_np.astype(np.float32).tofile(output_path)
+```
+Examples:
 
+```shell
```
ZwwWayne, line 236: shell -> python

PR #69, opened 2020-08-19
Fix minor bugs of computing iou3d

Tai-Wang:
Fix minor bugs encountered when computing iou3d with gpu while the specified cuda id is not default 0

### DIFF
--- a/mmdet3d/ops/iou3d/iou3d_utils.py
+++ b/mmdet3d/ops/iou3d/iou3d_utils.py
@@ -33,7 +33,6 @@ def nms_gpu(boxes, scores, thresh):
     Returns:
         torch.Tensor: Remaining indices with scores in descending order.
     """
-    # areas = (x2 - x1) * (y2 - y1)
     order = scores.sort(0, descending=True)[1]
 
     boxes = boxes[order].contiguous()
@@ -54,7 +53,6 @@ def nms_normal_gpu(boxes, scores, thresh):
     Returns:
         torch.Tensor: Remaining indices with scores in descending order.
     """
-    # areas = (x2 - x1) * (y2 - y1)
     order = scores.sort(0, descending=True)[1]
 
     boxes = boxes[order].contiguous()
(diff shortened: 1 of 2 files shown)


mmdet3d/ops/iou3d/iou3d_utils.py
```diff
@@ -4,13 +4,15 @@
...
-    """
+    """Calculate boxes IoU in the bird view.
+
+    Args:
+        boxes_a (torch.Tensor): Input boxes a with shape (M, 5).
+        boxes_b (torch.Tensor): Input boxes b with shape (N, 5).
 
+    Returns:
+        ans_iou (torch.Tensor): IoU result with shape (M, N).
+    """
```
ZwwWayne, line 15: 1. We may also fix this API by setting device.
2. ans_iou = boxes_a.new_zeros(torch.Size((boxes_a.shape[0], boxes_b.shape[0])))

mmdet3d/ops/iou3d/iou3d_utils.py
```diff
@@ -21,34 +23,41 @@ def boxes_iou_bev(boxes_a, boxes_b):
...
-    :param scores: (N)
-    :param thresh:
-    :return:
-    """
-    # areas = (x2 - x1) * (y2 - y1)
     order = scores.sort(0, descending=True)[1]
 
     boxes = boxes[order].contiguous()
 
     keep = torch.LongTensor(boxes.size(0))
```
ZwwWayne, line 40: Should we also create keep by `boxes.new_zeros(boxes.size(0))` ?

Tai-Wang: Yes. These ways to create new tensors could guarantee the normative access to specific cuda devices. (I tried to modify these parts before but I finally find the error is only caused by the cpp file. As a result I forget them in this PR...)

ZwwWayne (approved):

PR #78, opened 2020-08-24
Fix a minor bug in json2csv of lyft dataset

Tai-Wang:
This bug seems to be fixed before but was rolled back for some reasons.

### DIFF
--- a/mmdet3d/datasets/lyft_dataset.py
+++ b/mmdet3d/datasets/lyft_dataset.py
@@ -425,8 +425,7 @@ class LyftDataset(Custom3DDataset):
             pred_bboxes[..., 2] += pred_bboxes[..., 5] / 2
             show_result(points, gt_bboxes, pred_bboxes, out_dir, file_name)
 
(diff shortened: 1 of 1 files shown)


ZwwWayne (approved):

mmdet3d/datasets/lyft_dataset.py
```diff
@@ -433,8 +433,7 @@ def json2csv(json_path, csv_savepath):
             json_path (str): Path of the result json file.
             csv_savepath (str): Path to save the csv file.
         """
-        with open(json_path, 'r') as f:
-            results = mmcv.load(f)['results']
+        results = mmcv.load(json_path)['results']
         csv_nopred = 'data/lyft/sample_submission.csv'
```
ZwwWayne, line 437: 1. We may use `osp.join(self.data_root, 'sample_submission.csv')`
2. csv_nopred -> sample_list_path

ZwwWayne (approved):

PR #82, opened 2020-08-29
Feature add pointnet2 msg

encore-zhou:
Notice:

1. Add backbone for pointnet2_sa_msg.
2. Add unittest for pointnet2_sa_msg.
2. Code related to FPS can be ignored, these should be same as another PR.


mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -0,0 +1,173 @@
...
+        pool_mod (str): Pool method ('max' or 'avg') for SA modules.
+        use_xyz (bool): Whether to use xyz as a part of features.
+        normalize_xyz (bool): Whether to normalize xyz with radii in
+            each SA module.
+    """
+
+    def __init__(self,
+                 in_channels,
+                 num_points=(2048, 1024, 512, 256),
+                 radius=(0.2, 0.4, 0.8, 1.2),
```
ZwwWayne, line 37: radius -> radii

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -0,0 +1,173 @@
...
+            cur_xyz, cur_features, cur_indices = self.SA_modules[i](
+                sa_xyz[i], sa_features[i])
+            cur_features = self.aggregation_mlps[i](cur_features)
+            sa_xyz.append(cur_xyz)
+            sa_features.append(cur_features)
+            sa_indices.append(
+                torch.gather(sa_indices[-1], 1, cur_indices.long()))
+
+        ret = dict(
+            sa_xyz=sa_xyz[-1],
```
ZwwWayne, line 170: Add an argument `out_indices` thus the users could choose which stage to return. An example is [here](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/backbones/resnet.py#L316)

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -0,0 +1,190 @@
...
+            sa_xyz.append(cur_xyz)
+            sa_features.append(cur_features)
+            sa_indices.append(
+                torch.gather(sa_indices[-1], 1, cur_indices.long()))
+            if i in self.out_indices:
+                out_sa_xyz.append(sa_xyz[-1])
+                out_sa_features.append(sa_features[-1])
+                out_sa_indices.append(sa_indices[-1])
+
+        if len(self.out_indices) == 1:
```
ZwwWayne, line 181: Line 181 - 185 is unnecessary. If `len(self.out_indices) == 1`, then out_sa_xyz will only contain one element and could still be sent out.

xavierwu95 (approved):

ZwwWayne (approved):

ZwwWayne (approved):

PR #94, opened 2020-09-08
[Fix]: fix documentation related to nuim

ZwwWayne:


configs/nuimages/README.md
```diff
@@ -14,8 +14,8 @@ We will support panoptic segmentation models in the future.
```
Tai-Wang, line 14: Line 7: The baseline results includes->include

Tai-Wang: Line 31: All the training hyper-parameters follows -> follow

Tai-Wang (approved):

PR #95, opened 2020-09-08
[Fix]: fix multi-batch test problem

### LINKED ISSUE #89: if isinstance(result[0], tuple): KeyError: 0
Screenshot from 2020-09-04 17-45-45
Screenshot from 2020-09-04 17-45-26

seems like after updating mmdetection2.4 version , it trigger the bug above

xavierwu95:
1.Fix #89 
2. fix unittest bug in test_part_aggregation_ROI_head.


ZwwWayne:
Make the unit test data `tests/test_samples/parta2_roihead_inputs.npz` smaller than 100KB to reduce the burden of downloading/cloning the repo.

Tai-Wang (commented):
The only thing maybe needed to be improve is the size of .npz file. Saving it as .bin file or using other data type (like reduce float32 to float16) may help?

ZwwWayne (approved):

PR #104, opened 2020-09-14
[feature] add 3dssd benchmark

encore-zhou:


configs/3dssd/README.md
```diff
@@ -0,0 +1,19 @@
...
+    year = {2020}
+}
+```
+
+## Results
+
+### KITTI
+|  Backbone   |Class| Lr schd | Mem (GB) | Inf time (fps) | mAP |Download |
+| :---------: | :-----: | :------: | :------------: | :----: |:----: | :------: |
+|    [PointNet2SAMSG](./3dssd_kitti-3d-car.py)| Car |AdamW 72e|4.7||78.35||
```
ZwwWayne, line 19: 1. AdamW is not necessary here.
2. Add an note section and describe the difference between our implementation and the original implementation, especially why the results are different.

ZwwWayne (approved):

PR #121, opened 2020-09-21
[Enhance] Reduce requirements if waymo dataset is not used

Tai-Wang:
Reduce requirements if waymo dataset is not used.
1. Delete the installation of waymo dataset devkit in the requirement/runtime.txt
2. Move waymo related import into functions to avoid import errors
3. Add hints when importing waymo related functions and classes
4. Remind users to install waymo_open_dataset package in the doc


mmdet3d/datasets/waymo_dataset.py
```diff
@@ -184,6 +183,8 @@ def format_results(self,
                                                   pklfile_prefix,
                                                   submission_prefix)
         if 'waymo' in data_format:
+            from ..core.evaluation.waymo_utils.prediction_kitti_to_waymo import \
+                KITTI2Waymo  # noqa
```
ZwwWayne, line 187: is noqa here necessary?

Tai-Wang: Yes. Otherwise flake and isort would have a conflict. (isort tends to put `import` on the previous line while it makes it too long to trigger an error of flake, so I use the # noqa.)

ZwwWayne (approved):

PR #124, opened 2020-09-22
[feature] 3DSSD support nuscenes datasets

encore-zhou:
1. Modify input preparation in 3DSSD detector;
2. Modify ssd3d_head to support volecity prediction;
3. Modify related data pipelines.

### DIFF
--- a/mmdet3d/models/dense_heads/ssd_3d_head.py
+++ b/mmdet3d/models/dense_heads/ssd_3d_head.py
@@ -56,7 +56,9 @@ class SSD3DHead(VoteHead):
                  dir_res_loss=None,
                  size_res_loss=None,
                  corner_loss=None,
-                 vote_loss=None):
+                 vote_loss=None,
+                 velocity_loss=None):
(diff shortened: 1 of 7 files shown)


mmdet3d/models/dense_heads/ssd_3d_head.py
```diff
@@ -56,7 +56,9 @@ def __init__(self,
                  dir_res_loss=None,
                  size_res_loss=None,
                  corner_loss=None,
-                 vote_loss=None):
+                 vote_loss=None,
+                 velocity_loss=None):
+        self.extra_reg_dim = 2 if velocity_loss is not None else 0
```
Tai-Wang, line 61: Can we take the velocity as one of our regression targets of bounding boxes? It is the typical approach in anchor-based methods.

meng-zha:
Hi, I would like to know when this feature will be ready. Or maybe I need to modify the files by myself if I want the 3dssd supporting nuscenes dataset.

PR #132, opened 2020-09-24
Support to train using FP16

ZwwWayne:


mmdet3d/ops/pointnet_modules/point_fp_module.py
```diff
@@ -34,6 +35,7 @@ def __init__(self,
                     conv_cfg=dict(type='Conv2d'),
                     norm_cfg=norm_cfg))
 
+    @force_fp32(apply_to=('target', 'source', 'target_feats', 'source_feats'))
```
hellock, line 38: `@force_fp32()`

mmdet3d/ops/norm.py
```diff
@@ -42,10 +43,14 @@ class NaiveSyncBatchNorm1d(nn.BatchNorm1d):
         It is slower than `nn.SyncBatchNorm`.
     """
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fp16_enabled = False
+
+    @force_fp32(out_fp16=True)
```
hellock, line 50: All BN layers will be forced to use fp32 and there is no need to add this line.

mmdet3d/ops/furthest_point_sample/points_sampler.py
```diff
@@ -59,7 +60,9 @@ def __init__(self,
         self.samplers = nn.ModuleList()
         for fps_mod in fps_mod_list:
             self.samplers.append(get_sampler_type(fps_mod)())
+        self.fp16_enabled = False
 
+    @force_fp32(apply_to=('points_xyz', 'features'))
```
hellock, line 65: `@force_fp32()`

mmdet3d/ops/norm.py
ZwwWayne: This should be kept because in high version Pytorch the input of customized norm layers is FP16 but will not be converted due to [this line](https://github.com/open-mmlab/mmcv/blob/master/mmcv/runner/fp16_utils.py#L275).

hellock (approved):

mmdet3d/models/dense_heads/anchor3d_head.py
```diff
@@ -211,39 +213,64 @@ def loss_single(self, cls_score, bbox_pred, dir_cls_preds, labels,
...
+                    pos_dir_targets,
+                    pos_dir_weights,
+                    avg_factor=num_total_samples)
+        else:
+            loss_bbox = pos_bbox_pred.sum()
+            if self.use_direction_classifier:
+                loss_dir = pos_dir_cls_preds.sum()
+        if torch.isnan(loss_bbox).any():
+            import pdb
+            pdb.set_trace()
```
hellock, line 272: This should be removed?

PR #138, opened 2020-09-30
[Enhance] Waymo evaluation and new pp results

Tai-Wang:
1. Add overall mAP computation in waymo evaluation
2. Update load_interval=1 pointpillars baseline results


configs/_base_/datasets/waymo-3d-car.py
```diff
@@ -0,0 +1,122 @@
...
+        data_root=data_root,
+        ann_file=data_root + 'waymo_infos_val.pkl',
+        split='training',
+        pipeline=test_pipeline,
+        modality=input_modality,
+        classes=class_names,
+        test_mode=True,
+        box_type_3d='LiDAR'))
+
+evaluation = dict(interval=24)
```
ZwwWayne, line 122: Should this line be put in the model's real training config?

PR #211, opened 2020-11-30
[Fix]: create_gt_database

meng-zha:
* Fix bugs for the create_gt_database of kitti

* Fix bugs for '__get_item__' of BasePoints

PR #221, opened 2020-12-08
[Fix]: create_gt_database of nuscenes

meng-zha:
* Fix the bug on the commit #211 for the create_gt_database in nuscenes


meng-zha:
It will be appreciated if you can test the create_data.py for all datasets. I'm sorry I have made a small bugs in commit #221.

ZwwWayne (approved):

Issue #227, opened 2020-12-11
LoadMultiViewImageFromFiles fails in data pipeline


SimonDoll:
Shape error in mmdet3d/datasets/pipelines/formating.py when loading multi view images.
```Line 50: img = np.ascontiguousarray(results['img'].transpose(2, 0, 1))```
For multiview images `results['img']` is of shape (height, width, 3 (rgb), 6 (multi view))
The transpose call fails with `ValueError: axes don't match array` due to the last axis.

**Reproduction**
1. What command or script did you run?
```
python tools/test.py config checkpoint --eval mAp
```
I added loading multi view images to the default nuscenes config (train and test pipeline):
```
dict(
        type='LoadMultiViewImageFromFiles'
    ),
```
**Environment**
sys.platform: linux
Python: 3.7.7 (default, May  7 2020, 21:25:33) [GCC 7.3.0]
CUDA available: True
GPU 0: GeForce RTX 2070 SUPER
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 10.1, V10.1.243
GCC: gcc (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
PyTorch: 1.6.0
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) Math Kernel Library Version 2020.0.1 Product Build 20200208 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v1.5.0 (Git Hash e2ac1fac44c5078ca927cb9b90e1b3066a0b2ed0)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 10.1
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_37,code=compute_37
  - CuDNN 7.6.3
  - Magma 2.5.2
[...]

PR #229, opened 2020-12-11
Fixes LoadMultiViewImageFromFiles

SimonDoll:
Fixes a bug that prevents the use of LoadMultiViewImageFromFiles in data loading pipelines, see #227 for a bug report.
Is the new order of ```img = (rgb x h x w x rgb x multi_imgs)``` the desired one?
Any help is appreciated, I'm quite new to the framework.

### DIFF
--- a/mmdet3d/datasets/pipelines/formating.py
+++ b/mmdet3d/datasets/pipelines/formating.py
@@ -47,7 +47,10 @@ class DefaultFormatBundle(object):
                 imgs = np.ascontiguousarray(np.stack(imgs, axis=0))
                 results['img'] = DC(to_tensor(imgs), stack=True)
             else:
-                img = np.ascontiguousarray(results['img'].transpose(2, 0, 1))
+                # move rgb dimension to front
+                # (h x w x rgb x multi_imgs) -> (rgb x h x w x rgb x multi_imgs)
+                # last dim is  optional
+                img = np.ascontiguousarray(np.moveaxis(results['img'], 2, 0))
                 results['img'] = DC(to_tensor(img), stack=True)
         for key in [
                 'proposals', 'gt_bboxes', 'gt_bboxes_ignore', 'gt_labels',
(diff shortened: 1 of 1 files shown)


ZwwWayne:
Hi @SimonDoll ,
Thanks for your kind PR. Would you like to add a unit test for the PR to test your case?

SimonDoll:
Hi @ZwwWayne, thanks for Your reply.
I can add a test case for this, but as far as I understand the "dataset" provided for test cases (tests/data/nuscenes/) does only provide lidar data and no camera images.
Do You have a script at hand to recreate this test dataset that I can modify to include camera images?
Is test_nuscene_dataset.py the correct location for the test?
Thanks for any help!

tojimahammatov:
Hi @SimonDoll , I also faced this problem and understood the reason after some debugging. However, I think that it should be fixed by editing  ` loading.py ` , in particular  ` LoadMultiViewImageFromFiles `  class, line number 44. This line is responsible for making image data for  ` results ` dict.  ` img `  is not an instance of  ` list `  there, but  ` numpy.ndarray ` .

np.stack creates 4-dimensional data, but in format.py it is checked against instance of `list` which is not correct. Maybe changing that  `if`  line is better option, for example something like:

` if isinstance(results['img'], list) or len(results['img'].shape)>3: `
        `      pass    # img has 4 dims => multi view`

But anyway, your change might effect datasets having a single image (e.g.  `Kitti`)

I think the best way is creating  `list`  variable for `results['img']`  in `LoadMultiViewImageFromFiles` class.

Wuziyi616:
This PR is fixed in #611

OpenMMLab-Assistant-007:
Hi
@SimonDoll 
We are grateful for your efforts in helping improve this open-source project during your personal time. 
Welcome to join OpenMMLab Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/UjgXkPWNqA
If you have a WeChat account，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）
Thank you again for your contribution❤

PR #231, opened 2020-12-13
fix indoor_eval in case of less classes in prediction

### LINKED ISSUE #230: possible bug in indoor_eval
Looks like [`eval_map_recall`](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/evaluation/indoor_eval.py#L163) function is incorrect in case of `gt.keys() != pred.keys()`.
For example if `gt.keys() == [2, 4]` and `pred.keys() == [4]` then `len(ret_values) == 1` and when `i == 1` we get `RuntimeError` in [`ret_values[i]`](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/evaluation/indoor_eval.py#L194).
If you also think it is a bug, i can make a pull request to fix it.

filaPro:
Fixes #230.

### DIFF
--- a/mmdet3d/core/evaluation/indoor_eval.py
+++ b/mmdet3d/core/evaluation/indoor_eval.py
@@ -178,20 +178,20 @@ def eval_map_recall(pred, gt, ovthresh=None):
         tuple[dict]: dict results of recall, AP, and precision for all classes.
     """
 
-    ret_values = []
+    ret_values = {}
     for classname in gt.keys():
         if classname in pred:
-            ret_values.append(
-                eval_det_cls(pred[classname], gt[classname], ovthresh))
+            ret_values[classname] = eval_det_cls(
+                pred[classname], gt[classname], ovthresh)
     recall = [{} for i in ovthresh]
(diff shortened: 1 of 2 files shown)


Tai-Wang (commented):
Please notice messages from linting. You can pre-commit to check whether there is any problem in terms of format.

filaPro:
Hi @Tai-Wang,

I've fixed yapf warnings. However I have one more question. What is the idea behind `origin=(0.5, 0.5, 0.5)` [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/evaluation/indoor_eval.py#L257)? Looks like this function deals with `gt` and `pred` in terms of `DepthInstance3DBoxes` with default `origin == (0.5, 0.5, 0.0)`. But in this line you only shift `gt` along `z` axis and don't modify `z` coordinate of `pred`. Is this the expected behavour?

Tai-Wang (approved):

ZwwWayne (approved):

PR #246, opened 2020-12-24
Fix bug when num_features != 4

Sundrops:


ZwwWayne:
Fixed with unit test in https://github.com/open-mmlab/mmdetection3d/pull/275

PR #258, opened 2021-01-06
[Fix] fix bugs for 3DSSD triggered by empty GT

encore-zhou:


mmdet3d/datasets/custom_3d.py
```diff
@@ -105,7 +105,7 @@ def get_data_info(self, index):
         if not self.test_mode:
             annos = self.get_ann_info(index)
             input_dict['ann_info'] = annos
-            if self.filter_empty_gt and len(annos['gt_bboxes_3d']) == 0:
+            if self.filter_empty_gt and ~(annos['gt_labels_3d'] != -1).any():
```
ZwwWayne, line 108: Can we check which case will cause the issue that make `gt_labels_3d=-1`? This bug should be fixed in other places.

mmdet3d/models/dense_heads/ssd_3d_head.py
```diff
@@ -335,6 +335,30 @@ def get_targets_single(self,
         valid_gt = gt_labels_3d != -1
         gt_bboxes_3d = gt_bboxes_3d[valid_gt]
         gt_labels_3d = gt_labels_3d[valid_gt]
+
+        # Generate fake GT for empty scene
+        if valid_gt.sum() == 0:
+            vote_targets = points.new_zeros(self.num_candidates, 3)
+            center_targets = points.new_zeros(self.num_candidates, 3)
+            size_res_targets = points.new_zeros(self.num_candidates, 3)
+            dir_class_targets = points.new_zeros(
```
ZwwWayne, line 344: May add unit test for the case of empty GT.

mmdet3d/datasets/custom_3d.py
encore-zhou: Class label of removed objects will be set as -1 in [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/kitti_dataset.py#L177) . But we need to keep these Bbox for collision test.

ZwwWayne (approved):

Tai-Wang (approved):

ZwwWayne (approved):

Issue #305, opened 2021-02-10
Question on points_in_boxes_gpu


Divadi:
I have been trying to use this function, more generally the 'points_in_boxes' function for 3d boxes objects.
I tried a simple example, creating a grid of points from -10~9.8 in each coordinate, and a box centered at the origin:

Input:
```
from mmdet3d.ops.roiaware_pool3d import points_in_boxes_gpu
points = np.stack(np.meshgrid(
    np.arange(-10, 10, 0.2), 
    np.arange(-10, 10, 0.2), 
    np.arange(-10, 10, 0.2)), axis=0).reshape(3, -1).T.astype(np.float32)[None, ...]
print(points.shape)
print(points)
```
Output:
```
(1, 1000000, 3)
[[[-10.  -10.  -10. ]
  [-10.  -10.   -9.8]
  [-10.  -10.   -9.6]
  ...
  [  9.8   9.8   9.4]
  [  9.8   9.8   9.6]
  [  9.8   9.8   9.8]]]
```
Input:
```
res = points_in_boxes_gpu(
    torch.tensor(points, device='cuda:1'), 
    torch.tensor([0, 0, 0, 4, 4, 4, 0], device='cuda:1').float().unsqueeze(0).unsqueeze(0)
)
(res != -1).sum()
```
Output:
```
tensor(0, device='cuda:1')
```

Which then indicates that there are no points inside the box, however, I feel like there should be a lot of points in the box I created - am I using this function incorrectly? Thank you for your help.

ZwwWayne:
This is wierd. Please @THU17cyz have a look.

PR #311, opened 2021-02-18
[Fix] Fix issue #305

THU17cyz:
The bug mentioned in issue #305 is caused by a discordance between the current cuda device and the device of the input tensors. This may potentially cause unexpected behaviours in other cuda ops.


ZwwWayne:
Can we use the pytorch API https://pytorch.org/docs/stable/cuda.html#torch.cuda.set_device rather than modifying the CUDA code?

PR #322, opened 2021-02-27
[Refactor] VoteNet refactor

THU17cyz:
Modified backbone feature extraction in order that ImVoteNet can directly use VoteHead.


THU17cyz:
Made sure the performance was unaffected

ZwwWayne (approved):

PR #323, opened 2021-02-27
[Fix] Rewarp open3d

xiliu8006:
Reduce the dependence of open3d


ZwwWayne (approved):

mmdet3d/core/visualizer/show_result.py
```diff
@@ -81,6 +79,12 @@ def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=True):
         filename (str): Filename of the current frame.
         show (bool): Visualize the results online.
     """
+    try:
```
ZwwWayne, line 82: This place do not need to do that as the warning should be bumped in open3d_vis.py

ZwwWayne (approved):

Issue #320, opened 2021-02-25
Resolve pytorch warnings of nonzeros


ZwwWayne:
**Describe the feature**
Resolve pytorch warnings of nonzeros
**Motivation**
In pytorch 1.6, the code reports many warnings of deprecated usages of nonzeros during training.

Tai-Wang:
Fixed by #330.

PR #328, opened 2021-03-01
fix nonzero warning with as_tuple, fix requirements for open3d

wHao-Wu:
[Fix]: Fix warning of deprecated usages of nonzero during training with pytorch 1.6 (#320 )

* fix warning caused by upgrade of pytorch with specifying as_tuple=Fasle

[Fix]: Fix import error after installing latest open3d

* latest open3d only work in Ubuntu 18.04 because of the lack of GLIBC_2.27 in Ubuntu 16.04


requirements/optional.txt
```diff
@@ -1,2 +1,2 @@
-open3d
+open3d==0.9.0.0
```
ZwwWayne, line 1: Add comments here to indicate the reason.

PR #331, opened 2021-03-03
[Fix] Fix import error after installing latest open3d

wHao-Wu:
[Fix]: Fix import error after installing latest open3d

* The latest open3d need the support of file 'GLIBC_2.27', which only exist in Ubuntu 18.04, not in Ubuntu 16.04

### DIFF
--- a/requirements/optional.txt
+++ b/requirements/optional.txt
@@ -1,2 +1,3 @@
-open3d
+open3d==0.9.0.0
+# The latest open3d need the support of file 'GLIBC_2.27', which only exist in Ubuntu 18.04, not in Ubuntu 16.04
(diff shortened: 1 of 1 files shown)


requirements/optional.txt
```diff
@@ -1,2 +1,3 @@
-open3d
+open3d==0.9.0.0
+# The latest open3d need the support of file 'GLIBC_2.27', which only exist in Ubuntu 18.04, not in Ubuntu 16.04
```
ZwwWayne, line 2: `The latest open3d need 'GLIBC_2.27' that only exists in Ubuntu 18.04, not in Ubuntu 16.04`

PR #343, opened 2021-03-08
Refactor train.py/test.py

wHao-Wu:
[Enhance] Refactoring the train.py/test.py

* Changing details in train.py
  + Re-seting gpu_ids with distributed training mode
  + Adding "cfg-options" and deprecating "options" in arguments
  + Adding support for "custom_imports"

* Changing details in test.py
  + Changing the import path of test.py
    + Switching wrap_fp16_model from path "mmcv" to path "mmdet"
  + Adding support for "custom_imports"
  + Adding "cfg-options" and "eval-options", and deprecating "options" in argument
  + Adding support when test dataset is concatenated
  + Changing op when args.fuse_conv_bn is true
  + Changing interface in evaluation

* Verifying refactor above with SECOND model

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -2,13 +2,10 @@ from __future__ import division
 
 import argparse
 import copy
+import logging
+import mmcv
 import os
 import time
-import logging
-import warnings
-from os import path as osp
-
-import mmcv
 import torch
 import warnings
 from mmcv import Config, DictAction
--- a/tools/train.py
+++ b/tools/train.py
@@ -2,10 +2,13 @@ from __future__ import division
 
 import argparse
 import copy
-import logging
-import mmcv
 import os
 import time
+import logging
+import warnings
+from os import path as osp
+
+import mmcv
 import torch
 import warnings
 from mmcv import Config, DictAction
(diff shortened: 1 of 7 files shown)


tools/train.py
```diff
@@ -40,13 +42,28 @@ def parse_args():
         nargs='+',
         help='ids of gpus to use '
         '(only applicable to non-distributed training)')
-    parser.add_argument('--seed', type=int, default=0, help='random seed')
+    parser.add_argument('--seed', type=int, default=None, help='random seed')
```
ZwwWayne, line 45: keep it as 0

tools/train.py
```diff
@@ -150,8 +186,7 @@ def main():
         # save mmdet version, config file content and class names in
         # checkpoints as meta data
         cfg.checkpoint_config.meta = dict(
-            mmdet_version=__version__,
-            config=cfg.pretty_text,
+            mmdet_version=__version__ + get_git_hash()[:7],
```
ZwwWayne, line 189: This will only get the git hash of mmdet3d rather than mmdet.

wHao-Wu: So we only need to get the version of mmdet? Is it correct to remove the get_git_hash() because the get_git_hash() only returns the git hash of the current repo of mmdet3d?

ZwwWayne (approved):

PR #370, opened 2021-03-20
[Fix] Fix a bug when adding ('val', 1) to workflow

THU17cyz:
Val in workflow is intended to show the losses on test set. This pr fixes a bug when adding val to workflow.

Still, val uses train-time data augmentations, which is not preferred. However, one can slightly modify the config files and train.py in order to use test-time data augmentations for val. Therefore we do not address the issue in this pr. Perhaps some refactors might be considered in the future regarding this issue (the test_mode meaning and pipeline overriding here are also a bit confusing)?

### DIFF
--- a/tools/train.py
+++ b/tools/train.py
@@ -179,11 +179,11 @@ def main():
     datasets = [build_dataset(cfg.data.train)]
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
-        try:
-            val_dataset.pipeline = cfg.data.train.pipeline
         # in case we use a dataset wrapper
-        except AttributeError:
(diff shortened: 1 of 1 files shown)


tools/train.py
```diff
@@ -179,7 +179,15 @@ def main():
     datasets = [build_dataset(cfg.data.train)]
     if len(cfg.workflow) == 2:
         val_dataset = copy.deepcopy(cfg.data.val)
-        val_dataset.pipeline = cfg.data.train.pipeline
+        try:
```
ZwwWayne, line 182: Maybe we can directly check whether `dataset` is in `cfg.data.train`.

Tai-Wang (approved):

ZwwWayne (approved):

PR #379, opened 2021-03-24
[Enhance] Add numba and numpy version requirements in FAQ

xiliu8006:
Add numba and numpy version requirements in FAQ


docs/faq.md
```diff
@@ -16,3 +16,9 @@ We list some potential troubles encountered by users and developers, along with
   pip uninstall pycocotools mmpycocotools
   pip install mmpycocotools
   ```
+
+- If you face the error shown below when your numba == 0.48.0 and numpy >= 1.20.0
```
ZwwWayne, line 20: If you face the error shown below, and your environment contains numba == 0.48.0 with numpy >= 1.20.0

docs/faq.md
```diff
@@ -16,3 +16,9 @@ We list some potential troubles encountered by users and developers, along with
...
+
+  ``TypeError: expected dtype object, got 'numpy.dtype[bool_]'``
+
+  please downgrade numpy to 1.18.0 or install numba == 0.48 from source, because numpy 1.20.0 'np.dtype' now produces subclass. This is an API change.
```
ZwwWayne, line 24: Add a command to do that.

Tai-Wang (approved):

docs/faq.md
```diff
@@ -16,3 +16,14 @@ We list some potential troubles encountered by users and developers, along with
   pip uninstall pycocotools mmpycocotools
   pip install mmpycocotools
   ```
+
+- If you face the error shown below, and your environment contains numba == 0.48.0 with numpy >= 1.20.0
+
+  ``TypeError: expected dtype object, got 'numpy.dtype[bool_]'``
+
+  please downgrade numpy to 1.18.0 or install numba == 0.48 from source, because numpy 1.20.0 'np.dtype' now produces subclass. This is an API change. The current workaround is as below.
```
ZwwWayne, line 24: The numpy version must be 1.18.0? Could it be 1.19.0?

xiliu8006: it can be 1.19.0.  so should I change 1.18.0 to 1.19.0?  The reason I written 1.18.0 was that it could match this [issue](https://github.com/numba/numba/issues/6041) in numba

ZwwWayne: Then we should not force the numpy version, rather, we should give a valid range for that. Further, if numpy >=1.20.0, is there any numba version greater than 0.48.0 can fix the issue?

xiliu8006: if we want to use the numba version greater than 0.48.0, we need to modify some code. So we can only choose one workaround.

ZwwWayne (approved):

PR #416, opened 2021-04-07
[Feature] Support online visualization of semantic segmentation results

Wuziyi616:
Similar to detection online visualization, I use open3d to show segmentation masks. The main difference here is that, in det, we can draw gt and pred bbox on top of scene_points, because they won't overlap too much with each other. However the drawing of seg requires per-point colorization, while scene_points often have their own colors. So we can't draw them together.

My solution is that, I set an offset for each seg mask and then draw them in a row. See examples below.

### DIFF
--- a/mmdet3d/core/visualizer/open3d_vis.py
+++ b/mmdet3d/core/visualizer/open3d_vis.py
@@ -1,3 +1,4 @@
+import copy
 import cv2
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


Wuziyi616:
The saved offline results (on ScanNet dataset, the pred_mask is randomly generated):

Points:
pts

GT:
gt

Pred:
pred

Online visualization:
all

Tai-Wang (approved):

ZwwWayne (approved):

Issue #411, opened 2021-04-06
Errors in docstrings of pre-processing functions


Wuziyi616:
The input points of pre-processing functions in ```pipelines/``` are instances of BasePoints class, but the comments say they are np.ndarray. Needs to fix the comment errors.

PR #422, opened 2021-04-07
[Fix] Fix point type errors in docstring in pipeline functions

Wuziyi616:
Fix [issue#411](https://github.com/open-mmlab/mmdetection3d/issues/411), so that developers will understand the points during loading pipeline are instances of BasePoints class. They need to take out the tensor/coord/attribute of points in case some operations are not supported by BasePoints class.

### DIFF
--- a/mmdet3d/datasets/pipelines/loading.py
+++ b/mmdet3d/datasets/pipelines/loading.py
@@ -154,7 +154,7 @@ class LoadPointsFromMultiSweeps(object):
         """Removes point too close within a certain radius from origin.
 
         Args:
-            points (np.ndarray): Sweep points.
+            points (np.ndarray | :obj:`BasePoints`): Sweep points.
             radius (float): Radius below which points are removed.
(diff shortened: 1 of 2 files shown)


mmdet3d/datasets/pipelines/loading.py
```diff
@@ -154,7 +154,7 @@ def _remove_close(self, points, radius=1.0):
         """Removes point too close within a certain radius from origin.
 
         Args:
-            points (np.ndarray): Sweep points.
+            points (np.ndarray | BasePoints): Sweep points.
```
Tai-Wang, line 157: :obj:`BasePoints`

Tai-Wang (approved):

mmdet3d/datasets/pipelines/loading.py
Wuziyi616: Fix.

ZwwWayne (approved):

Issue #412, opened 2021-04-06
Data loading in show and evaluate functions of dataset class


Wuziyi616:
Data loading issue when data is on ceph rather than disk, where we can't use np to load them. Consider re-using self.pipeline to load them.

PR #430, opened 2021-04-09
[Enhance] Add pipeline for data loading

Wuziyi616:
Fix [issue#412](https://github.com/open-mmlab/mmdetection3d/issues/412).

In dataset.show/evaluate() functions, we may need to load points/gt. In current implementations, we load them from disk using ```np.fromfile()```, which is incompatible when data is in ceph. To solve this, we add ```eval_pipeline``` in configs, and pass it as an argument to these functions. This pipeline purely consists of raw data loading operations (e.g. LoadImage, LoadPoints), eliminating the effects of data augmentation, and can adjust with the file client.

### DIFF
--- a/configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
+++ b/configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
@@ -133,32 +133,8 @@ test_pipeline = [
             dict(type='Collect3D', keys=['points'])
         ])
 ]
-# construct a pipeline for data and gt loading in show function
-# please keep its loading function consistent with test_pipeline (e.g. client)
-eval_pipeline = [
-    dict(
-        type='LoadPointsFromFile',
-        coord_type='LIDAR',
-        load_dim=5,
-        use_dim=5,
-        file_client_args=file_client_args),
-    dict(
-        type='LoadPointsFromMultiSweeps',
-        sweeps_num=9,
-        use_dim=[0, 1, 2, 3, 4],
-        file_client_args=file_client_args,
-        pad_empty_sweeps=True,
-        remove_close=True),
-    dict(
-        type='DefaultFormatBundle3D',
-        class_names=class_names,
-        with_label=False),
-    dict(type='Collect3D', keys=['points'])
-]
 
 data = dict(
     train=dict(dataset=dict(pipeline=train_pipeline)),
     val=dict(pipeline=test_pipeline),
     test=dict(pipeline=test_pipeline))
-
-evaluation = dict(interval=20, pipeline=eval_pipeline)
--- a/configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
+++ b/configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
@@ -133,8 +133,32 @@ test_pipeline = [
             dict(type='Collect3D', keys=['points'])
         ])
 ]
+# construct a pipeline for data and gt loading in show function
+# please keep its loading function consistent with test_pipeline (e.g. client)
+eval_pipeline = [
+    dict(
+        type='LoadPointsFromFile',
+        coord_type='LIDAR',
+        load_dim=5,
+        use_dim=5,
+        file_client_args=file_client_args),
+    dict(
+        type='LoadPointsFromMultiSweeps',
+        sweeps_num=9,
+        use_dim=[0, 1, 2, 3, 4],
+        file_client_args=file_client_args,
+        pad_empty_sweeps=True,
+        remove_close=True),
+    dict(
+        type='DefaultFormatBundle3D',
+        class_names=class_names,
+        with_label=False),
+    dict(type='Collect3D', keys=['points'])
+]
 
 data = dict(
     train=dict(dataset=dict(pipeline=train_pipeline)),
     val=dict(pipeline=test_pipeline),
     test=dict(pipeline=test_pipeline))
+
+evaluation = dict(interval=20, pipeline=eval_pipeline)
--- a/mmdet3d/datasets/custom_3d.py
+++ b/mmdet3d/datasets/custom_3d.py
@@ -226,7 +226,8 @@ class Custom3DDataset(Dataset):
                  iou_thr=(0.25, 0.5),
                  logger=None,
                  show=False,
-                 out_dir=None):
+                 out_dir=None,
+                 pipeline=None):
         """Evaluate.
 
         Evaluation in indoor protocol.
@@ -238,6 +239,8 @@ class Custom3DDataset(Dataset):
             show (bool): Whether to visualize.
                 Default: False.
             out_dir (str): Path to save the visualization results.
+                Default: None.
+            pipeline (list[dict], optional): raw data loading for showing.
                 Default: None.
 
         Returns:
@@ -262,9 +265,28 @@ class Custom3DDataset(Dataset):
             box_type_3d=self.box_type_3d,
             box_mode_3d=self.box_mode_3d)
         if show:
-            self.show(results, out_dir)
+            self.show(results, out_dir, pipeline=pipeline)
 
         return ret_dict
+
+    @staticmethod
+    def _extract_data(results, key):
+        """Extract and return the data corresponding to key in results dict.
+
+        Args:
+            results (dict): Result dict containing point clouds data.
+            key (str): Key of the desired data.
+
+        Returns:
+            np.ndarray | torch.Tensor: Data term.
+        """
+        # results[key] may be DataContainer or list[DataContainer]
+        if isinstance(results[key], mmcv.parallel.DataContainer):
+            data = results[key]._data
+        else:  # list output of MultiScaleFlipAug3D
+            data = results[key][0]._data
+
+        return data
 
     def __len__(self):
         """Return the length of data infos.
--- a/mmdet3d/datasets/custom_3d.py
+++ b/mmdet3d/datasets/custom_3d.py
@@ -1,12 +1,14 @@
 import mmcv
 import numpy as np
 import tempfile
+import warnings
 from os import path as osp
 from torch.utils.data import Dataset
 
 from mmdet.datasets import DATASETS
 from ..core.bbox import get_box_type
 from .pipelines import Compose
+from .utils import get_loading_pipeline
 
 
 @DATASETS.register_module()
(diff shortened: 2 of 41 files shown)


Wuziyi616:
The added ```eval_pipeline``` will be passed as an argument in the [eval_hook](https://github.com/open-mmlab/mmdetection/blob/master/mmdet/core/evaluation/eval_hooks.py#L176-L177).

**Work in progress**, loading for ScanNet-seg dataset need to be done. For configs using different data pipeline from ```_base_/datasets/```, I also need to modify them...

Wuziyi616:
Actually the commit message shouldn't be "reuse", because I pass into a new pipeline. "Reuse" is for easy understanding since the added eval_pipeline is similar to test_pipeline.

mmdet3d/datasets/custom_3d_seg.py
```diff
@@ -354,22 +357,46 @@ def evaluate(self,
...
         ), f'Expect elements in results to be dict, got {type(results[0])}.'
+
+        # we need to load mask annotation so set test_mode as False
+        original_test_mode = self.test_mode
+        self.test_mode = False
+        if pipeline is not None:
+            pipeline = Compose(pipeline)
+            original_pipeline = self.pipeline if \
+                hasattr(self, 'pipeline') else None  # save the original one
+            self.pipeline = pipeline  # set new pipeline for data loading
```
Tai-Wang, line 368: Must we change the original self.pipeline and self.test_mode?

mmdet3d/datasets/custom_3d.py
```diff
@@ -262,10 +265,31 @@ def evaluate(self,
...
+        Returns:
+            np.ndarray | torch.Tensor: Data term.
+        """
+        # results[key] may be data or list[data]
+        # data may be wrapped inside DataContainer
+        data = results[key]
+        if isinstance(data, list) or isinstance(data, tuple):
+            data = data[0]
+        if isinstance(data, mmcv.parallel.DataContainer):
+            data = data._data
```
Tai-Wang, line 289: Maybe add an else will be more safe and clear.

mmdet3d/datasets/custom_3d_seg.py
Wuziyi616: Good question. Usually we call ```evaluate()``` and ```show()``` in eval or test, where ```self.pipeline``` is often ```test_pipeline``` (in ```config.py```). As you know, ```test_pipeline``` often involves pre-processing transforms of points/img (e.g. ```PointRangeFilter```, ```Normalize```) and doesn't have ```LoadAnnotation``` functions. Therefore, we need to change the original ```self.pipeline```.

As for ```self.test_mode```, it's ```True``` in eval and test, so dataset won't load annotation information (see [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/custom_3d_seg.py#L123-L125)). Therefore, we need to set it as ```False``` to load the path of GT seg mask.

Essentially, the difference stems from the GT in det and seg task. In det, gt bbox is stored in ```self.data_infos``` (see [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/scannet_dataset.py#L83-L86)), so we don't need ```LoadAnnotation3D``` function in ```eval_pipeline```. But GT seg mask is not stored in ```self.data_infos``` and could only be loaded from file, so we need to set the pipeline and test_mode.

mmdet3d/datasets/custom_3d.py
Wuziyi616: I don't know, what should I do in else? I have taken ```data``` as ```result[key]```, even if the two ```if``` doesn't hit, we still have assigned value to ```data```. What do you think? Some advice please.

mmdet3d/datasets/scannet_dataset.py
```diff
@@ -107,26 +108,42 @@ def get_ann_info(self, index):
...
             results (list[dict]): List of bounding boxes results.
             out_dir (str): Output directory of visualization result.
             show (bool): Visualize the results online.
+            pipeline (list[dict], optional): raw data loading for showing.
+                Default: None.
         """
         assert out_dir is not None, 'Expect out_dir, got none.'
+        if pipeline is not None:
+            pipeline = Compose(pipeline)
+            original_pipeline = self.pipeline if \
```
ZwwWayne, line 124: Should not use self.pipeline because this may override the original `self.pipeline` that might be used in other places, e.g., the dataset is used for evaluate model after show.

Wuziyi616:
Move the tedious if/else conditions to dataset._extract_data(). Set default value to pipeline.

Wuziyi616:
I have tried using ```tools/misc/visualize_results.py``` to show results by calling dataset.show() and input config.eval_pipeline. The results are the same as before.

Wuziyi616:
Basic logic now:
- if pipeline is given in show/evaluate function, directly use this pipeline
- if pipeline is None, if self.pipeline is not None, ```get_loading_pipeline(self.pipeline)```
- pipeline and self.pipeline are both None, call ```_build_default_pipeline()``` for this dataset

mmdet3d/datasets/custom_3d.py
```diff
@@ -262,10 +267,83 @@ def evaluate(self,
...
+        """
+        assert pipeline is not None, 'data loading pipeline is not provided'
+        if load_annos:
+            original_test_mode = self.test_mode
+            self.test_mode = False
+        input_dict = self.get_data_info(index)
+        self.pre_pipeline(input_dict)
+        example = pipeline(input_dict)
+
+        def get_data(key):
```
Tai-Wang, line 318: Make it as a static method in the class? Defining functions in the function is a little confusing and unclear.

mmdet3d/datasets/custom_3d_seg.py
```diff
@@ -382,6 +372,79 @@ def _rand_another(self, idx):
...
+        """
+        assert pipeline is not None, 'data loading pipeline is not provided'
+        if load_annos:
+            original_test_mode = self.test_mode
+            self.test_mode = False
+        input_dict = self.get_data_info(index)
+        self.pre_pipeline(input_dict)
+        example = pipeline(input_dict)
+
+        def get_data(key):
```
Tai-Wang, line 419: Also make it as a static method

mmdet3d/datasets/kitti_dataset.py
```diff
@@ -670,22 +674,61 @@ def convert_valid_bboxes(self, box_dict, info):
...
+                    coord_type='LIDAR',
+                    load_dim=4,
+                    use_dim=4,
+                    file_client_args=dict(backend='disk')),
+                dict(
+                    type='DefaultFormatBundle3D',
+                    class_names=self.CLASSES,
+                    with_label=False),
+                dict(type='Collect3D', keys=['points'])
+            ]
```
Tai-Wang, line 707: Simplify these two cases? First define the pipeline with a list including the shared 3 pipelines, then insert dict(type='LoadImageFromFile') for different cases.

mmdet3d/datasets/utils.py
```diff
@@ -1,21 +1,59 @@
...
+    Returns:
+        bool | None: Whether it is a loading function. None means can't judge.
+            When transform is `MultiScaleFlipAug3D`, we return None.
+    """
+    # TODO: use more elegant way to distinguish loading modules
+    loading_functions = (LoadImageFromFile, LoadPointsFromFile,
+                         LoadAnnotations3D, LoadMultiViewImageFromFiles,
+                         LoadPointsFromMultiSweeps, DefaultFormatBundle3D,
+                         Collect3D, LoadImageFromFileMono3D,
+                         PointSegClassMapping)
```
Tai-Wang, line 30: A little dangerous. We may not remember to add new defined loading functions to here when implementing a new one. Is there any other good way? Or at least we need to move them into __init__ or somewhere else such that it is convenient for us to remember to update them when adding new functions?

mmdet3d/datasets/custom_3d.py
Wuziyi616: Done.

mmdet3d/datasets/custom_3d_seg.py
Wuziyi616: Done.

mmdet3d/datasets/kitti_dataset.py
Wuziyi616: Good idea.

mmdet3d/datasets/utils.py
Wuziyi616: Do you think it's ok if we regard all functions defined in `formating.py` and `loading.py` as loading functions? I have checked and this rule applies to all functions except one (`NormalizePointsColor`, which I do think should be placed in `transform_3d.py` instead of `loading.py`). What do you think @ZwwWayne?

configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
```diff
@@ -133,8 +133,32 @@
...
+        with_label=False),
+    dict(type='Collect3D', keys=['points'])
+]
 
 data = dict(
     train=dict(dataset=dict(pipeline=train_pipeline)),
     val=dict(pipeline=test_pipeline),
     test=dict(pipeline=test_pipeline))
+
+evaluation = dict(interval=20, pipeline=eval_pipeline)
```
ZwwWayne, line 164: Since the configs support inheritance. We can only modify base config of datasets and do not need to modify so many configs?

mmdet3d/datasets/custom_3d_seg.py
```diff
@@ -382,6 +372,82 @@ def _rand_another(self, idx):
...
+            key (str | list[str]): One single or a list of data key.
+            load_annos (bool): Whether to load data annotations.
+                If True, need to set self.test_mode as False before loading.
+
+        Returns:
+            np.ndarray | torch.Tensor | list[np.ndarray | torch.Tensor]:
+                A single or a list of loaded data.
+        """
+        assert pipeline is not None, 'data loading pipeline is not provided'
+        if load_annos:
```
ZwwWayne, line 434: indicate which part needs `test_mode` to be True in the comments.

Wuziyi616 (commented):
I have carefully checked all the config files and 5 of them needed not to be modified.

configs/centerpoint/centerpoint_0075voxel_second_secfpn_4x8_cyclic_20e_nus.py
Wuziyi616: Yes, for sure.

mmdet3d/datasets/utils.py
Wuziyi616: I asked Wenwei and he said currently we can just leave it as is. Maybe make it better in the future.

mmdet3d/datasets/custom_3d_seg.py
Wuziyi616: Done.

ZwwWayne:
PRs can be merged after resolving conflicts.

PR #439, opened 2021-04-13
[Enhance] Move ScanNet point alignment from data pre-processing to pipeline

Wuziyi616:
In the original ScanNet pre-processing script for the detection task, the point cloud scene is aligned according to axis_align_matrix in the annotation (see [here](https://github.com/open-mmlab/mmdetection3d/blob/master/data/scannet/load_scannet_data.py#L89-L103)). However, ScanNet test set doesn't have such alignment matrix, so people don't do such alignment in the semantic segmentation task (because they need to submit the test set results to online benchmark, while det only test and reports on val set). In order for ScanNet seg and set to share one pre-processed data, we decided to move the alignment step to pipeline.


Wuziyi616:
- Modify ScanNet pre-processing scripts to save the axis_align_matrix to data_info.pkl
- Support Box.rotate function when input is **angle** or **rotation_matrix**, add unittest for it

mmdet3d/core/bbox/structures/cam_box3d.py
```diff
@@ -180,10 +182,19 @@ def rotate(self, angle, points=None):
         """
         if not isinstance(angle, torch.Tensor):
             angle = self.tensor.new_tensor(angle)
-        rot_sin = torch.sin(angle)
-        rot_cos = torch.cos(angle)
-        rot_mat_T = self.tensor.new_tensor([[rot_cos, 0, -rot_sin], [0, 1, 0],
-                                            [rot_sin, 0, rot_cos]])
+        assert angle.shape == torch.Size([3, 3]) or angle.numel() == 1
```
Tai-Wang, line 185: assertion message

mmdet3d/core/bbox/structures/depth_box3d.py
```diff
@@ -125,11 +127,20 @@ def rotate(self, angle, points=None):
         """
         if not isinstance(angle, torch.Tensor):
             angle = self.tensor.new_tensor(angle)
-        rot_sin = torch.sin(angle)
-        rot_cos = torch.cos(angle)
-        rot_mat_T = self.tensor.new_tensor([[rot_cos, -rot_sin, 0],
-                                            [rot_sin, rot_cos, 0], [0, 0,
-                                                                    1]]).T
+        assert angle.shape == torch.Size([3, 3]) or angle.numel() == 1
```
Tai-Wang, line 130: assertion message

mmdet3d/core/bbox/structures/cam_box3d.py
Wuziyi616: Fix.

mmdet3d/core/bbox/structures/depth_box3d.py
Wuziyi616: Add.

Wuziyi616:
After discussing with Wenwei, we decided to extract **both** aligned and unaligned_bbox during pre-processing of ScanNet dataset. For the aligned bbox, they are stored in info with the **same** key as before (e.g. `gt_boxes_upright_depth`, `dimension`), while for the unaligned bbox, they are stored with `unaligned_` marked keys (e.g. `unaligned_gt_boxes_upright_depth`).

As aligned bbox are stored with the same keys, we can keep most of the code unchanged (e.g. `dataset.evaluate()`). If we would like to support 3D instance segmentation on ScanNet in the future, we will need the `unaligned_bbox` infos.

Wuziyi616:
I have checked the correctness of processed data by testing a pre-trained VoteNet on it and got mAP_25, mAP_50 close to the originally reported values (<1%). I think the data is OK to use.

Wuziyi616 (commented):
I think this PR is ready for review. @Tai-Wang @ZwwWayne Thanks.

tools/data_converter/scannet_data_utils.py
```diff
@@ -197,9 +219,6 @@ def _convert_to_label(self, mask):
                 mask = np.load(mask)
             else:
                 mask = np.fromfile(mask, dtype=np.long)
-        # first filter out unannotated points (labeled as 0)
-        mask = mask[mask != 0]
-        # then convert to [0, 20) labels
```
Wuziyi616, line 202: Remove these lines to be consistent with data pre-processing steps.

Wuziyi616:
I have trained a VoteNet on the new processed ScanNet data. The results are as follows:

| Data | AP@0.25 | AP@0.5 |
| :-----: | :----: | :----: |
| Old | 62.90 | 39.91 |
| New | 62.75 | 40.07 |

I think this result is very close to original training behavior.

Tai-Wang (approved):

data/scannet/load_scannet_data.py
```diff
@@ -52,6 +52,29 @@ def read_segmentation(filename):
...
+def extract_bbox(mesh_vertices, object_id_to_segs, object_id_to_label_id,
+                 instance_ids):
+    num_instances = len(np.unique(list(object_id_to_segs.keys())))
+    instance_bboxes = np.zeros((num_instances, 7))
+    for obj_id in object_id_to_segs:
+        label_id = object_id_to_label_id[obj_id]
+        obj_pc = mesh_vertices[instance_ids == obj_id, 0:3]
+        if len(obj_pc) == 0:
+            continue
+        xmin = np.min(obj_pc[:, 0])
```
ZwwWayne, line 64: Can we use the following to simplify this part?
```python
xyz_min = np.min(obj_pc, axis=-1)
xyz_max = np.max(obj_pc, axis=-1)
bbox = np.concatenate([(xyz_min+xyz_max)/2, xyz_max - xyz_min], axis=-1)
```

mmdet3d/core/bbox/structures/depth_box3d.py
```diff
@@ -130,11 +132,21 @@ def rotate(self, angle, points=None):
...
+            f'invalid rotation angle shape {angle.shape}'
+
+        if angle.numel() == 1:
+            rot_sin = torch.sin(angle)
+            rot_cos = torch.cos(angle)
+            rot_mat_T = self.tensor.new_tensor([[rot_cos, -rot_sin, 0],
+                                                [rot_sin, rot_cos, 0],
+                                                [0, 0, 1]]).T
+        else:
+            rot_mat_T = angle.T
```
ZwwWayne, line 145: We should assert/check the shape of `rot_mat_T` and raise value error for safety.

PR #441, opened 2021-04-14
[Fix] Waymo loading error

wHao-Wu:
* Fix waymo config when loading point


Tai-Wang (approved):

Wuziyi616 (approved):
I have checked this change in the added Waymo unit test. It works well.

ZwwWayne (approved):

PR #442, opened 2021-04-14
[Feature] Support FCOS3D head

Tai-Wang:


mmdet3d/models/dense_heads/anchor_free_mono3d_head.py
```diff
@@ -0,0 +1,514 @@
...
+            self.conv_dir_cls = nn.Conv2d(self.dir_branch[-1], 2, 1)
+        if self.pred_attrs:
+            self.conv_attr_prev = self._init_branch(
+                conv_channels=self.attr_branch,
+                conv_strides=(1, ) * len(self.attr_branch))
+            self.conv_attr = nn.Conv2d(self.attr_branch[-1], self.num_attrs, 1)
+
+    def init_weights(self):
+        """Initialize weights of the head."""
+        for m in self.cls_convs:
```
ZwwWayne, line 269: Can we simplify the logic, like
```python
for modules in [self.cls_convs, self.reg_convs, self.conv_cls_prev]:
    for m in modules:
        xxxxx

```

mmdet3d/models/dense_heads/anchor_free_mono3d_head.py
```diff
@@ -0,0 +1,514 @@
...
+            tuple: Scores for each class, bbox predictions, direction class,
+                and attributes, features after classification and regression
+                conv layers, some models needs these features like FCOS.
+        """
+        cls_feat = x
+        reg_feat = x
+
+        for cls_layer in self.cls_convs:
+            cls_feat = cls_layer(cls_feat)
+        clone_cls_feat = cls_feat.clone()
```
ZwwWayne, line 342: Do we really need to clone feature here?

mmdet3d/models/dense_heads/anchor_free_mono3d_head.py
```diff
@@ -0,0 +1,514 @@
...
+        for conv_cls_prev_layer in self.conv_cls_prev:
+            clone_cls_feat = conv_cls_prev_layer(clone_cls_feat)
+        cls_score = self.conv_cls(clone_cls_feat)
+
+        for reg_layer in self.reg_convs:
+            reg_feat = reg_layer(reg_feat)
+        bbox_pred = []
+        for i in range(len(self.group_reg_dims)):
+            clone_reg_feat = reg_feat.clone()
```
ZwwWayne, line 351: seems clone_reg_feat are cloned multiple times

mmdet3d/models/dense_heads/anchor_free_mono3d_head.py
Tai-Wang: Yes. Both `cls_feat` and `reg_feat` are need to be cloned for reusing them afterwards, otherwise it will be modified when applying convolutions in different branches.

Tai-Wang: I have added comments when it is necessary to clone feature maps.

ZwwWayne (approved):

Issue #440, opened 2021-04-13
Pipeline functions may not process points and masks simultaneously


Wuziyi616:
In [PointShuffle](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/pipelines/transforms_3d.py#L446), we only shuffle the order of points. Do we also need to change the order of other labels, e.g. semantic_mask and instance_mask accordingly?

Wuziyi616:
Not only PointShuffle, for functions that filter points (e.g. PointsRangeFilter), we don't process mask accordingly in the current implementation.

PR #444, opened 2021-04-15
[Enhance] Pipeline function process points and masks simultaneously

Wuziyi616:
Fix [issue#440](https://github.com/open-mmlab/mmdetection3d/issues/440). When filtering/sampling points in pipeline, perform the same actions on masks too.

### DIFF
--- a/mmdet3d/core/bbox/structures/box_3d_mode.py
+++ b/mmdet3d/core/bbox/structures/box_3d_mode.py
@@ -64,19 +64,19 @@ class Box3DMode(IntEnum):
         """Convert boxes from `src` mode to `dst` mode.
 
         Args:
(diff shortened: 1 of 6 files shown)


Tai-Wang (approved):

Wuziyi616:
Maybe one potential question about `ObjectSample` in `transforms_3d.py`. If I understand correctly, this function sub-samples GTs so that we can train on a sub-class of original all classes. As it performs sampling, we also need to consider seg_mask here (currently these is only gt_bbox processing in this function). However, when I look into `dbsampler` and it seems impossible to fix this issue in a few lines of code. I tend to leave this function into future PRs.

ZwwWayne (approved):

PR #453, opened 2021-04-16
[Fix] Fix error in four unittest functions

THU17cyz:


THU17cyz:
See the changed files. Three errors originate from the unittests and one originate from a config file which was missed in the config file refactor pr.

ZwwWayne (approved):

PR #466, opened 2021-04-21
added mmocr link

quincylin1:
Added mmocr link to chinese and english readme


ZwwWayne:
This PR involves irrelevant changes, can we clean them and only modify the two readmes?

innerlee:
@quincylin1 the diff is not right. Pls check again

PR #473, opened 2021-04-22
[Fix] enable visualization of the demo results online

wHao-Wu:
* Enabling visualization of the pcd demo results online
* Enabling visualization of the multi-modality demo results online

**Risk**: Whether to simultaneously visualize the point and image results online when visualizing multi-modality demo?

### DIFF
--- a/mmdet3d/core/visualizer/show_result.py
+++ b/mmdet3d/core/visualizer/show_result.py
@@ -68,7 +68,7 @@ def _write_oriented_bbox(scene_bbox, out_filename):
     return
 
 
-def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=True):
+def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=False):
     """Convert results into format that is directly readable for meshlab.
 
     Args:
@@ -77,8 +77,11 @@ def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=True):
         pred_bboxes (np.ndarray): Predicted boxes.
         out_dir (str): Path of output directory
         filename (str): Filename of the current frame.
-        show (bool): Visualize the results online. Defaults to True.
+        show (bool): Visualize the results online. Defaults to False.
     """
+    result_path = osp.join(out_dir, filename)
+    mmcv.mkdir_or_exist(result_path)
+
     if show:
         from .open3d_vis import Visualizer
 
@@ -87,10 +90,8 @@ def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=True):
             vis.add_bboxes(bbox3d=pred_bboxes)
         if gt_bboxes is not None:
             vis.add_bboxes(bbox3d=gt_bboxes, bbox_color=(0, 0, 1))
-        vis.show()
-
-    result_path = osp.join(out_dir, filename)
-    mmcv.mkdir_or_exist(result_path)
+        show_path = osp.join(result_path, f'{filename}_online.png')
+        vis.show(show_path)
 
     if points is not None:
         _write_obj(points, osp.join(result_path, f'{filename}_points.obj'))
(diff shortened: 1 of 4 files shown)


wHao-Wu:
## PCD demo
### SECOND:
kitti_000008_online

### VoteNet:
sunrgbd_000017_online

## Multi-modality demo
### MVX:
kitti_000008_online

### ImVoteNet:
sunrgbd_000017_online

Wuziyi616:
@wHao-Wu Are we able to visualize img and points at the same time? As far as I know, we need to press `q` to exit open3d.vis and mmcv.imshow(). How can you show them simultaneously? I know there is a so-called non-blocking mode in opencv but I haven't tried that before.

BTW, in multi-modality browse_dataset, I first show 3D point results, and then show 2D img results when the user exits 3D window.

mmdet3d/core/visualizer/show_result.py
```diff
@@ -87,10 +90,8 @@ def show_result(points, gt_bboxes, pred_bboxes, out_dir, filename, show=True):
             vis.add_bboxes(bbox3d=pred_bboxes)
         if gt_bboxes is not None:
             vis.add_bboxes(bbox3d=gt_bboxes, bbox_color=(0, 0, 1))
-        vis.show()
-
-    result_path = osp.join(out_dir, filename)
-    mmcv.mkdir_or_exist(result_path)
+        show_path = osp.join(result_path, f'{filename}_online.png')
+        vis.show(show_path)
```
Wuziyi616, line 94: I don't think we should save the online results every time, at least let's pass a parameter to determine whether to save? Like what I've done [here](https://github.com/Wuziyi616/mmdetection3d/blob/multi_dataset_browse/mmdet3d/core/visualizer/show_result.py#L102-105)?

wHao-Wu: Thanks for your advice. I have also considered the problem of mandatory saving online results. I will modify the corresponding operation lately.

wHao-Wu:
> @wHao-Wu Are we able to visualize img and points at the same time? As far as I know, we need to press `q` to exit open3d.vis and mmcv.imshow(). How can you show them simultaneously? I know there is a so-called non-blocking mode in opencv but I haven't tried that before.
> 
> BTW, in multi-modality browse_dataset, I first show 3D point results, and then show 2D img results when the user exits 3D window.

wHao-Wu:
> @wHao-Wu Are we able to visualize img and points at the same time? As far as I know, we need to press `q` to exit open3d.vis and mmcv.imshow(). How can you show them simultaneously? I know there is a so-called non-blocking mode in opencv but I haven't tried that before.
> 
> BTW, in multi-modality browse_dataset, I first show 3D point results, and then show 2D img results when the user exits 3D window.

I am sorry I have not explained the problem clearly. I mean that maybe we don't need to visualize the points results every time because we have shown the final results in which the detected boxes have been drawn in images.

Wuziyi616:
> > @wHao-Wu Are we able to visualize img and points at the same time? As far as I know, we need to press `q` to exit open3d.vis and mmcv.imshow(). How can you show them simultaneously? I know there is a so-called non-blocking mode in opencv but I haven't tried that before.
> > BTW, in multi-modality browse_dataset, I first show 3D point results, and then show 2D img results when the user exits 3D window.
> 
> I am sorry I have not explained the problem clearly. I mean that maybe we don't need to visualize the points results every time because we have shown the final results in which the detected boxes have been drawn in images.

Maybe we can ask what @ZwwWayne thinks. In my opinion, both 3d points and 2d image are input data so it would be better if we show both?

baheytharwat:
> @wHao-Wu Are we able to visualize img and points at the same time? As far as I know, we need to press `q` to exit open3d.vis and mmcv.imshow(). How can you show them simultaneously? I know there is a so-called non-blocking mode in opencv but I haven't tried that before.
> 
> BTW, in multi-modality browse_dataset, I first show 3D point results, and then show 2D img results when the user exits 3D window.

@Wuziyi616 , Is there a way to update the open3d visualization window without pressing `q` each time? I am using `show_result_meshlab` function from the demo example
I just want to display the detection results of multiple point clouds sequentially without pressing any key

PR #479, opened 2021-04-24
[Feature] Support PointNet++ decode head

Wuziyi616:
Add decode_head into codebase for segmentation task. Implement a base head and a PN++ head.


.github/workflows/build.yml
```diff
@@ -92,6 +92,7 @@ jobs:
         run: |
           pip install mmcv-full -f https://download.openmmlab.com/mmcv/dist/cu101/${{matrix.torch_version}}/index.html
           pip install -q git+https://github.com/open-mmlab/mmdetection.git
+          pip install -q git+https://github.com/open-mmlab/mmsegmentation.git
```
Wuziyi616, line 95: Our decode head and loss inherits from MMSeg.

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -102,15 +102,21 @@ def __init__(self,
...
-                    sa_out_channel,
-                    aggregation_channels[sa_index],
-                    conv_cfg=dict(type='Conv1d'),
-                    norm_cfg=dict(type='BN1d'),
-                    kernel_size=1,
-                    bias=True))
-            sa_in_channel = aggregation_channels[sa_index]
+
+            cur_aggregation_channel = aggregation_channels[sa_index]
+            if cur_aggregation_channel is None:
```
Wuziyi616, line 107: In the original implementation of PN++ for segmentation, we don't have this aggregation MLP. So I judge if it's None here.

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -139,14 +145,15 @@ def forward(self, points):
         sa_features = [features]
         sa_indices = [indices]
 
-        out_sa_xyz = []
-        out_sa_features = []
-        out_sa_indices = []
+        out_sa_xyz = [xyz]
```
Wuziyi616, line 148: In PN++ for segmentation, we need all the `sa_xyz` and `sa_features`, so I include the first xyz here.
I have checked the codebase, only 3DSSD uses PN2MSG, and it only uses the **last** `out_sa_xyz` and `out_sa_features`, so adding a first element won't affect its performance.

mmdet3d/models/backbones/pointnet2_sa_ssg.py
```diff
@@ -132,5 +132,10 @@ def forward(self, points):
             fp_indices.append(sa_indices[self.num_sa - i - 1])
 
         ret = dict(
-            fp_xyz=fp_xyz, fp_features=fp_features, fp_indices=fp_indices)
+            fp_xyz=fp_xyz,
+            fp_features=fp_features,
+            fp_indices=fp_indices,
+            sa_xyz=sa_xyz,
```
Wuziyi616, line 138: We need not only `fp` but also `sa` xyz/features in PN++ for seg.

mmdet3d/models/decode_heads/decode_head.py
```diff
@@ -0,0 +1,118 @@
+from abc import ABCMeta, abstractmethod
+from mmcv.cnn import normal_init
+from mmcv.runner import auto_fp16, force_fp32
+from torch import nn as nn
+
+from mmseg.models.builder import build_loss
```
Wuziyi616, line 6: We need `CrossEntropy` loss from MMSeg but not MMDet (it's somehow strange for me that they use different implementations).

Wuziyi616:
I have added some comments about the changes I made above. Please check them before reviewing my code :)

mmdet3d/models/decode_heads/decode_head.py
Tai-Wang: Do you think there will be many places that we need to use something from MMSeg? We should try our best to avoid more dependencies. I just want to make sure it is a necessity in the future.
If necessary, maybe we can make it an optional requirement when installing mmdet3d, and also please add corresponding instructions in the documentation.

Wuziyi616: Very good problem. In this head PR it seems unnecessary, but in the Segmentor3D class, I inherited a lot from Segmentor2D. Besides, we need to use the `eval_hook` from MMSeg so that we evaluate mIoU rather than mAP.

Tai-Wang (approved):

Wuziyi616:
Due to some mistakes in my work, I would like to commit the whole PN2 model in one PR (I really did some modifications on PN2 `decode_head` when working on PN2 `Segmentor`), and the benchmark results in another PR. I will leave this PR open and let you review it in advance, but I will raise another PR later.

Wuziyi616 (commented):
@ZwwWayne I think this PR is ready.

ZwwWayne (approved):

PR #481, opened 2021-04-24
[Fix] Recycle gpu memory in unit tests

THU17cyz:
After adding FCOS3D model, running all unit tests will result in a gpu out-of-memory error. Recycling gpu memory in every single model test can resolve the problem. Please review the changes I made and see whether this change should be applied to other unit tests or not, and whether there is a better solution.

### DIFF
--- a/tests/test_models/test_detectors.py
+++ b/tests/test_models/test_detectors.py
@@ -75,14 +75,17 @@ def test_get_dynamic_voxelnet():
     if not torch.cuda.is_available():
         pytest.skip('test requires GPU and torch+cuda')
 
-    dynamic_voxelnet_cfg = _get_model_cfg(
-        'dynamic_voxelization/dv_second_secfpn_6x8_80e_kitti-3d-car.py')
-    self = build_detector(dynamic_voxelnet_cfg).cuda()
-    points_0 = torch.rand([2010, 4], device='cuda')
-    points_1 = torch.rand([2020, 4], device='cuda')
-    points = [points_0, points_1]
-    feats = self.extract_feat(points, None)
-    assert feats[0].shape == torch.Size([2, 512, 200, 176])
+    with torch.no_grad():
+        dynamic_voxelnet_cfg = _get_model_cfg(
+            'dynamic_voxelization/dv_second_secfpn_6x8_80e_kitti-3d-car.py')
+        self = build_detector(dynamic_voxelnet_cfg).cuda()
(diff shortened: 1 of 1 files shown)


THU17cyz:
Running `PYTHONPATH=./ srun --partition xxx --ntasks=1 --gres=gpu:1 --ntasks-per-node=1 pytest -s tests` gives us:

`RuntimeError: CUDA out of memory. Tried to allocate 20.00 MiB (GPU 0; 11.90 GiB total capacity; 7.49 GiB already allocated; 10.94 MiB free; 7.65 GiB reserved in total by PyTorch`

Tai-Wang (approved):

tests/test_models/test_detectors.py
```diff
@@ -75,14 +75,17 @@ def test_get_dynamic_voxelnet():
...
-    feats = self.extract_feat(points, None)
-    assert feats[0].shape == torch.Size([2, 512, 200, 176])
+    with torch.no_grad():
+        dynamic_voxelnet_cfg = _get_model_cfg(
+            'dynamic_voxelization/dv_second_secfpn_6x8_80e_kitti-3d-car.py')
+        self = build_detector(dynamic_voxelnet_cfg).cuda()
+        points_0 = torch.rand([2010, 4], device='cuda')
+        points_1 = torch.rand([2020, 4], device='cuda')
+        points = [points_0, points_1]
+        feats = self.extract_feat(points, None)
```
ZwwWayne, line 85: we should use no_grad just above this line, rather than a big code block

tests/test_models/test_detectors.py
```diff
@@ -92,74 +95,81 @@ def test_voxel_net():
...
+        gt_bboxes = [gt_bbox_0, gt_bbox_1]
+        gt_labels_0 = torch.randint(0, 3, [10], device='cuda')
+        gt_labels_1 = torch.randint(0, 3, [10], device='cuda')
+        gt_labels = [gt_labels_0, gt_labels_1]
+        img_meta_0 = dict(box_type_3d=LiDARInstance3DBoxes)
+        img_meta_1 = dict(box_type_3d=LiDARInstance3DBoxes)
+        img_metas = [img_meta_0, img_meta_1]
+
+        # test forward_train
+        losses = self.forward_train(points, img_metas, gt_bboxes, gt_labels)
```
ZwwWayne, line 114: only forward need no_grad

PR #483, opened 2021-04-25
[Fix] Put simple_test under torch.no_grad() in unit test

THU17cyz:
This solves the oom problem in unit test correctly and should replace #481 . Shall we put every `simple_test` under the `test_models` folder to be under `torch.no_grad()`?

### DIFF
--- a/tests/test_models/test_detectors.py
+++ b/tests/test_models/test_detectors.py
@@ -113,7 +113,8 @@ def test_voxel_net():
     assert losses['loss_dir'][0] >= 0
 
     # test simple_test
-    results = self.simple_test(points, img_metas)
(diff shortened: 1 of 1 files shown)


Wuziyi616 (approved):
I agree, I also encountered this problem in testing my 3D seg models. It's very reasonable I think.

Tai-Wang (approved):

ZwwWayne (approved):

ZwwWayne:
> This solves the oom problem in unit test correctly and should replace #481 . Shall we put every `simple_test` under the `test_models` folder to be under `torch.no_grad()`?

I agree. If the change is a little big, we can merge this PR first and do that in our next version.

PR #489, opened 2021-04-26
[Enhance] Efficient implementation of PointSegClassMapping

Wuziyi616:
The original implementation uses for-loop to convert every label, which is super slow in large scenes (e.g. S3DIS seg dataset without sub-sampling). I modify it to use matrix indexing, which could be faster. Matrix indexing is also a common practice in other 3D Seg repos.

### DIFF
--- a/mmdet3d/datasets/pipelines/loading.py
+++ b/mmdet3d/datasets/pipelines/loading.py
@@ -242,12 +242,16 @@ class PointSegClassMapping(object):
     """
 
     def __init__(self, valid_cat_ids, max_cat_id=40):
+        assert max_cat_id >= np.max(valid_cat_ids), \
+            'max_cat_id should be greater than maximum id in valid_cat_ids'
+
         self.valid_cat_ids = valid_cat_ids
-        self.max_cat_id = max_cat_id
+        self.max_cat_id = int(max_cat_id)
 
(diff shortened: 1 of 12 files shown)


Wuziyi616:
I have carefully checked and modified all the places which use `PointSegClassMapping` in the codebase.

mmdet3d/datasets/pipelines/loading.py
```diff
@@ -237,10 +237,19 @@ class PointSegClassMapping(object):
...
     Args:
         valid_cat_ids (tuple[int]): A tuple of valid category.
+        max_cat_id (int): The max possible cat_id in input segmentation mask.
+            Defaults to 40.
     """
 
-    def __init__(self, valid_cat_ids):
+    def __init__(self, valid_cat_ids, max_cat_id=40):
         self.valid_cat_ids = valid_cat_ids
+        self.max_cat_id = max_cat_id
```
ZwwWayne, line 246: Need to check consistency between max_cat_id and valid_cat_id

ZwwWayne: also test the check in unittests

Wuziyi616: Done.

ZwwWayne (approved):

PR #495, opened 2021-04-27
[Refactor] Use MMCV MODEL_REGISTRY

Wuziyi616:
Note: this requires MMCV>=1.3.1!

PR #497, opened 2021-04-28
[Fix] Typos in .gitignore & error in customize_runtime docs

Wuziyi616:
The ignored data folder should be `data/xxx` instead of `/data/xxx`.
MMDet no longer has `mmdet/core/optimizer`. Instead, they use optimizer from `mmcv`. I modify this part of docs accordingly.

### DIFF
--- a/docs/tutorials/customize_runtime.md
+++ b/docs/tutorials/customize_runtime.md
@@ -20,8 +20,8 @@ To modify the learning rate of the model, the users only need to modify the `lr`
 A customized optimizer could be defined as following.
 
 Assume you want to add a optimizer named `MyOptimizer`, which has arguments `a`, `b`, and `c`.
-You need to create a new directory named `mmdet/core/optimizer`.
-And then implement the new optimizer in a file, e.g., in `mmdet/core/optimizer/my_optimizer.py`:
+You need to create a new directory named `mmdet3d/core/optimizer`.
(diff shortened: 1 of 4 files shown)


docs/tutorials/customize_runtime.md
```diff
@@ -24,7 +24,7 @@ You need to create a new directory named `mmdet/core/optimizer`.
 And then implement the new optimizer in a file, e.g., in `mmdet/core/optimizer/my_optimizer.py`:
```
Wuziyi616, line 24: Where do you think is better to add the file? If we add at `mmdet`, we assume the user installed `mmdet ` via git clone. This is the same if we add at `mmcv`. I would prefer to add this file `mmdet3d/core/optimizer/` which is the easiest to implement.

Wuziyi616:
This issue may exist in all downstream repos because `optimizer` is now implemented in `mmcv`. At least the customize_runtime docs of MMDet and MMSeg are both wrong.

Wuziyi616:
The CI error is because MMDet has updated to use MMCV's MODEL registry. This check can be passed after merging #495.

docs/tutorials/customize_runtime.md
ZwwWayne: should add to mmdet3d since mmdet is taken as a lib by default

ZwwWayne (approved):

PR #501, opened 2021-04-28
[Fix] Fix init_weight problem by reverting PR#495 and requiring mmdet<=2.11.0

Wuziyi616:
Wait to see the CI here. It works well on my PC.


Tai-Wang (approved):

Wuziyi616:
Next time we need to:
1. Update the mmdet and mmcv version requirement
2. Modify mmdet3d/models/build.py to use mmcv.MODEL registry

Wuziyi616:
Also need to modify the workflow I think...

PR #502, opened 2021-04-29
[Enhance] Support MMCLI

wHao-Wu:
* Support auto-installation in MMCLI


requirements/mminstall.txt
```diff
@@ -0,0 +1,2 @@
+mmcv-full 1.2.4 1.4.0
+mmdet 2.10.0 3.0.0
```
ZwwWayne, line 2: mmdet 2.10.0 2.11.0

wHao-Wu: Thanks for your advice. I have updated the version of mmdet in mminstall.txt.

ZwwWayne (approved):

PR #470, opened 2021-04-21
[Enhance] Add compatibility document

Wuziyi616:
I imitate the compatibility document in [MMDet](https://github.com/open-mmlab/mmdetection/blob/master/docs/compatibility.md) and write a (very initial and very naive) version. Only includes the change on ScanNet dataset.

PR #163, opened 2020-10-12
[Fix]: add convert script for Votenet

encore-zhou:
Fix issue #145 by  converting the keys in the checkpoints of votenet models.

PR #504, opened 2021-04-29
[Doc] Add compatibility doc based on #470

THU17cyz:
Based on MMDetection [compatibility doc](https://github.com/open-mmlab/mmdetection/blob/master/docs/compatibility.md) and #470 .


docs/compatibility.md
```diff
@@ -0,0 +1,15 @@
+# Compatibility with Previous Versions of MMDetection3D
+
+This document provides detailed descriptions of the BC-breaking changes in MMDetection3D.
+
+## MMDetection3D 0.12.0
+
+### SUNRGBD dataset for ImVoteNet
```
Wuziyi616, line 7: I remember there is a BC-breaking regarding VoteNet model and its pre-trained weight. Should we also mention it here?

docs/compatibility.md
```diff
@@ -0,0 +1,15 @@
...
+
+We adopt a new pre-processing procedure for the SUNRGBD dataset in order to support ImVoteNet, which is a multi-modality method requiring both image and point cloud data. In previous versions of MMDetection3D, SUNRGBD dataset was only used for point cloud based 3D detection methods. In MMDetection3D 0.12.0, we add ImVoteNet to our model zoo, thus updating SUNRGBD correspondingly by adding image-related pre-processing steps. Specificly, we made these changes:
+
+- Fix a bug in the image file path in meta data.
+- Convert calibration matrices from double to float to avoid type mismatch in further operations.
+- Add instructions in the documents on preparing image data.
+
+Please refer to the SUNRGBD [README.md](../data/sunrgbd/README.md) for more details.
```
Wuziyi616, line 15: Should use absolute path instead of relative path because users may click it online? (I remember Wenwei @ZwwWayne told me this before, is that correct?)

Wuziyi616 (commented):
LGTM except for some minor comments.

docs/compatibility.md
ZwwWayne: Yes, we'd better use the link in readthedocs.

ZwwWayne (approved):

docs/compatibility.md
Tai-Wang: If you must use link to some files, please also use the absolute path, because when browsing the web-based documentation, the relative path can not be found.

docs/compatibility.md
THU17cyz: Can you point me to that pull request?

Wuziyi616: I am not sure about the PR link, but if you look at the VoteNet's page [here](https://github.com/open-mmlab/mmdetection3d/blob/master/configs/votenet/README.md) (**Result** section, **Notice** part), you can see instruction about converting pre-trained model weight.

Wuziyi616: This #163 seems related, @ZwwWayne what do you think?

THU17cyz: Yes, but isn't this too ancient?

Wuziyi616: Not sure if that's too ancient. At least it seems indeed a BC-breaking to me. And when I first use MMDet3D, I forgot to convert the pre-trained weight lol. It's okay if you don't add it.

THU17cyz: I agree. I'll add it :)

docs/compatibility.md
THU17cyz: then shall I change every relative path to absolute path in the docs?

docs/0_demo.md
```diff
@@ -2,7 +2,7 @@
 
 ## Introduction
 
-We provide scipts for multi-modality/single-modality and indoor/outdoor 3D detection demos. The pre-trained models can be downloaded from [model zoo](../docs/model_zoo.md). We provide pre-processed sample data from KITTI and SUN RGB-D dataset. You can use any other data following our pre-processing steps.
+We provide scipts for multi-modality/single-modality and indoor/outdoor 3D detection demos. The pre-trained models can be downloaded from [model zoo](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/model_zoo.md). We provide pre-processed sample data from KITTI and SUN RGB-D dataset. You can use any other data following our pre-processing steps.
```
THU17cyz, line 5: Also fixed the relative path here. No more relative paths under `docs` folder.

Wuziyi616 (approved):

README_zh-CN.md
```diff
@@ -118,6 +118,8 @@ MMDetection3D 是一个基于 PyTorch 的目标检测开源工具箱, 下一代
 
 请参考[快速入门文档](docs/get_started.md)学习 MMDetection3D 的基本使用。 我们为新手提供了分别针对[已有数据集](docs/1_exist_data_model.md)和[新数据集](docs/2_new_data_model.md)的使用指南。我们也提供了一些进阶教程，内容覆盖了[学习配置文件](docs/tutorials/config.md), [增加数据集支持](docs/tutorials/customize_dataset.md), [设计新的数据预处理流程](docs/tutorials/data_pipeline.md), [增加自定义模型](docs/tutorials/customize_models.md), [增加自定义的运行时配置](docs/tutorials/customize_runtime.md)和 [Waymo 数据集](docs/tutorials/waymo.md).
 
+请参考[FAQ](docs/faq.md)查看一些常见的问题与解答。在升级MMDetection3D的版本时，请查看[兼容性文档](docs/compatibility.md)以知晓每个版本引入的不与之前版本兼容的更新。
```
ZwwWayne, line 121: 1. 升级 MMDetection3D 的
2. 参考 FAQ 查

docs/compatibility.md
```diff
@@ -0,0 +1,21 @@
...
+
+This document provides detailed descriptions of the BC-breaking changes in MMDetection3D.
+
+## MMDetection3D 0.6.0
+
+### VoteNet model structure update
+
+In MMDetection 0.6.0, we updated the model structure of VoteNet, therefore model checkpoints generated by MMDetection < 0.6.0 should be first converted to a format compatible with the latest VoteNet structure via this [script](https://github.com/open-mmlab/mmdetection3d/blob/master/tools/model_converters/convert_votenet_checkpoints.py). For more details, please refer to the VoteNet [README.md](https://github.com/open-mmlab/mmdetection3d/tree/master/configs/votenet/README.md)
+
+## MMDetection3D 0.12.0
```
ZwwWayne, line 11: 0.12.0 should be shown before 0.6.0

ZwwWayne (approved):

PR #515, opened 2021-05-04
Fix: #473 changed default arg value for `show_result` and caused `--show` broke during testing

haotian-liu:
In #473, it [changed the default arg value of `mmdet3d.core.visualizer.show_result`](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/visualizer/show_result.py#L78), from `show=True` to `show=False`, which caused code `mmdet3d.models.detectors.{mvx_two_stage/base}.show_results` to broke (not showing the visualization even when `--show` is enabled during testing.

### DIFF
--- a/mmdet3d/core/visualizer/show_result.py
+++ b/mmdet3d/core/visualizer/show_result.py
@@ -75,7 +75,7 @@ def show_result(points,
                 pred_bboxes,
                 out_dir,
                 filename,
-                show=False,
+                show=True,
                 snapshot=False):
     """Convert results into format that is directly readable for meshlab.
 


wHao-Wu:
The `--show` in testing only controls the offline visualization and creates files which can be visualized by [MeshLab](http://www.meshlab.net/). But the [show](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/visualizer/show_result.py#L78) in visualizer controls the online visualization. So you can add arguments in testing to control whether to visualize online.

mmdet3d/core/visualizer/show_result.py
```diff
@@ -75,7 +75,7 @@ def show_result(points,
                 pred_bboxes,
                 out_dir,
                 filename,
-                show=False,
```
wHao-Wu, line 78: The --show in testing only controls the offline visualization and creates files which can be visualized by MeshLab. But the show in visualizer controls the online visualization. So you can add arguments in testing to control whether to visualize online.

haotian-liu: However, this actually made `--show` in `./tools/test.py` broke:
In test.py, it calls `single_gpu_test` with `args.show` as the argument, and when that is `True`, it will [call](https://github.com/open-mmlab/mmdetection3d/blob/b255d16b9f8a19a8f1065a1ae6c88fd2a03dc325/mmdet3d/apis/test.py#L37-L41) `model.module.show_results(...)`, where in the function it [calls](https://github.com/open-mmlab/mmdetection3d/blob/b255d16b9f8a19a8f1065a1ae6c88fd2a03dc325/mmdet3d/models/detectors/mvx_two_stage.py#L502) `mmdet3d.core.show_result` without `show` argument (which in all of previous versions, it is default to `True`).
In the newest version, the default value of `show` in `mmdet3d.core.show_result` is set to `False`, which makes all of the previous code broke, if we update to the new version (and this is not mentioned in the Changelog, and if other users have used it in different places assuming the original behavior of `show_result`, it will also broke).
Therefore, this default value should not be changed to `show=False`, and should remain `show=True` as previous versions.

wHao-Wu: Thanks for your advice. Next, we will incorporate the 3D Segmentation into the repo. So we want to split the online visualization and offline visualization separately. For now, we can set the default `show` to true in show_results, which may be modified because of the consistency to 3D Segmentation.

haotian-liu: Thanks! And at that time, maybe we should change the behavior of `model.module.show_results(...)` by adding the parameter `show=True` when calling `mmdet3d.core.show_result`.

wHao-Wu: You are kindly right about the behavior. And we will take the feature into account soon.

OpenMMLab-Coodinator:
Hi @haotian-liu ！First of all, we want to express our gratitude for your significant PR in this project. Your contribution is highly appreciated, and we are grateful for your efforts in helping improve this open-source project during your personal time. We believe that many developers will benefit from your PR.

We would also like to invite you to join our Special Interest Group (SIG) private channel on Discord, where you can share your experiences, ideas, and build connections with like-minded peers. To join the SIG channel, simply message moderator— OpenMMLab on Discord or briefly share your open-source contributions in the #introductions channel and we will assist you. Look forward to seeing you there! Join us ：https://discord.gg/raweFPmdzG

If you are Chinese or have WeChat，welcome to join our community on WeChat. You can add our assistant ：openmmlabwx. Please add "mmsig + Github ID" as a remark when adding friends：）
Thank you again for your contribution❤

PR #540, opened 2021-05-12
[Enhance] Support rotate points without bbox in GlobalRotScaleTrans

Wuziyi616:
In the current implementation of `GlobalRotScaleTrans`, we will rotate points only when bbox is presented in `input_dict` ([here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/pipelines/transforms_3d.py#L453-L458)). However, in PC Seg, we need to rotate points without bbox. In this PR, I add a `if` condition for this situation. I also add some input check & perform params transformation in `__init__` to save future time.


Wuziyi616:
I discover that in [this](https://github.com/open-mmlab/mmdetection3d/blob/master/tests/test_data/test_pipelines/test_augmentations/test_test_augment_utils.py#L54-L58) unit test, someone has tested `GlobalRotScaleTrans` and the `expected_points` don't perform rotation. So, I wonder if 'don't rotate points when bbox is not presented' is the desired behavior of this function?

Tai-Wang (approved):

Tai-Wang:
> I discover that in [this](https://github.com/open-mmlab/mmdetection3d/blob/master/tests/test_data/test_pipelines/test_augmentations/test_test_augment_utils.py#L54-L58) unit test, someone has tested `GlobalRotScaleTrans` and the `expected_points` don't perform rotation. So, I wonder if _don't rotate points when bbox is not presented_ is the desired behavior of this function?

I guess it should be a legacy issue. We don't have a separate class for `points` before, so previously some transformations related to points are achieved with bounding boxes (a hack workaround). Similar problems exist [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/pipelines/transforms_3d.py#L69). Maybe we can gradually refactor these similar operations in the future.

ZwwWayne (approved):

PR #546, opened 2021-05-14
[Fix] Add a file and instructions to fix the Lyft dataset

THU17cyz:


THU17cyz:
Shall this be included in compatibility or some other doc?

Tai-Wang (approved):

Tai-Wang:
> Shall this be included in compatibility or some other doc?

This change should not affect the name of the converted info file? If this only modifies that raw data, I think the current fix is ok enough.

THU17cyz:
> > Shall this be included in compatibility or some other doc?
> 
> This change should not affect the name of the converted info file? If this only modifies that raw data, I think the current fix is ok enough.

Yes, will not affect info file. I agree.

PR #547, opened 2021-05-14
[Fix] Updating mmcv_version to be consistent with mmseg and to support the mmcli

wHao-Wu:
* MMSeg needs mmcv=1.3.1 as the mmcv_minimum_version. Updating the mmcv_minimum_version in order to be compatible with MMSeg and therefore to support 3D Segmentation in MMDet3D and MMCLI.


Wuziyi616:
We also need to update the mmcv requirement [here](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/getting_started.md).

Tai-Wang (approved):

ZwwWayne (approved):

PR #548, opened 2021-05-15
[Fix] Change urls that end with .md under the `docs` folder

THU17cyz:


THU17cyz:
I just found a bug in read-the-docs. The read-the-docs generation strips `.md` out of every external link (or url) ending with `.md`. For example, [here](https://github.com/open-mmlab/mmdetection3d/blob/169a17afac9800898d103ab0933950626b259b79/docs/data_preparation.md#s3dis-scannet-and-sun-rgb-d) the link in "To prepare s3dis data, please see s3dis." will be converted to:

```
https://github.com/open-mmlab/mmdetection3d/blob/master/data/s3dis/README
```

in read-the-docs.

I searched for a solution, and the closest answer I could find is [this](https://github.com/sphinx-doc/sphinx/issues/6631). Therefore I provide an easy workaround. It seems that adding a slash after README.md will tell sphinx that this is an external link, thus solving the problem.

THU17cyz:
I have already tested this on my fork.

Wuziyi616 (commented):
This workaround can indeed fix the errors and I think it's acceptable. However, I discover that in MMDet's docs, although they write `.md`, they become `.html` in the generated read-the-docs. And their links can successfully navigate to the desired page. Can we find out why and also adopt it in MMDet3D?

Wuziyi616:
> This workaround can indeed fix the errors and I think it's acceptable. However, I discover that in MMDet's docs, although they write `.md`, they become `.html` in the generated read-the-docs. And their links can successfully navigate to the desired page. Can we find out why and also adopt it in MMDet3D?

It seems that's because `.md` files using relative path won't have this error and will be converted. Then I think this workaround is necessary.

Wuziyi616 (approved):

THU17cyz:
> > This workaround can indeed fix the errors and I think it's acceptable. However, I discover that in MMDet's docs, although they write `.md`, they become `.html` in the generated read-the-docs. And their links can successfully navigate to the desired page. Can we find out why and also adopt it in MMDet3D?
> 
> It seems that's because `.md` files using relative path won't have this error and will be converted. Then I think this workaround is necessary.

Yes, relative paths ending with `.md` are not affected.

THU17cyz:
And MMDetection read-the-docs has the same problem.

ZwwWayne:
> And MMDetection read-the-docs has the same problem.

can raise an issue.

Issue #449, opened 2021-04-15
Possible bug in nuscenes evaluation?


filaPro:
Is [this](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/nuscenes_dataset.py#L437) `if` correct?
Typical 3d detector returns `list(dict('boxes_3d':..., 'scores': ..., 'labels_3d': ...), ...)`. If i'm not missing something, in this case we want to go to the `if` branch here, not `else`.

Tai-Wang:
Actually we will give a key like 'pts_bbox' to each prediction (an example is [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/models/detectors/mvx_two_stage.py#L411)), so the result should be like `list(dict('pts_bbox': dict('boxes_3d':..., 'scores': ..., 'labels_3d': ...), 'img_bbox': dict('boxes_3d':..., 'scores': ..., 'labels_3d': ...)), ...)`. Hence, it should enter the `else`.

filaPro:
Yes, I saw it. But for `KITTI` detectors (ex. VoxelNet, SSD3D) the result doesn't have this `pts_bbox` and `img_bbox` keys, so they can not be used for `NuScenes` without modifying this `if` condition.

Tai-Wang:
OK I see. Maybe we should use a more compatible condition to distinguish these two cases, like `if not ('pts_bbox' in results[0] or 'img_bbox' in results[0])`(although it's a little hack). Do you have such cases? If yes, could you please help us fix and test it?

filaPro:
Yes, I'm experimenting with custom single stage detector. Your solution works fine. Can make a one line PR from it.

PR #549, opened 2021-05-15
[Fix] Fix evaluation bugs in Nus and Lyft datasets

Wuziyi616:
Fix issue #449

### DIFF
--- a/mmdet3d/datasets/lyft_dataset.py
+++ b/mmdet3d/datasets/lyft_dataset.py
@@ -335,9 +335,15 @@ class LyftDataset(Custom3DDataset):
         else:
             tmp_dir = None
 
-        if not isinstance(results[0], dict):
+        # this is a bit hack to enable testing KITTI detectors on nuScenes
+        # refer to https://github.com/open-mmlab/mmdetection3d/issues/449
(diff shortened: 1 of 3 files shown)


mmdet3d/datasets/lyft_dataset.py
```diff
@@ -335,9 +335,15 @@ def format_results(self, results, jsonfile_prefix=None, csv_savepath=None):
         else:
             tmp_dir = None
 
-        if not isinstance(results[0], dict):
+        # this is a bit hack to enable testing KITTI detectors on nuScenes
```
Tai-Wang, line 338: It's not necessarily KITTI detectors. The issue just takes those models as examples. We can just add a general comment here to indicate the expected format of results.

Tai-Wang: Also please note that here should be Lyft instead of nuScenes

Wuziyi616: I am not sure if my modification is acceptable now. Please kindly check it.

Tai-Wang (approved):

ZwwWayne (approved):

PR #556, opened 2021-05-18
[Fix] Converting point with specified rt_matrix

wHao-Wu:
* Fixing the bug that the specified rt_matrix will be overwritten by fixed constant when converting from lidar to cam, or from cam to lidar, or from lidar to depth, or from depth to lidar.

### DIFF
--- a/mmdet3d/core/bbox/structures/coord_3d_mode.py
+++ b/mmdet3d/core/bbox/structures/coord_3d_mode.py
@@ -220,9 +220,11 @@ class Coord3DMode(IntEnum):
         # convert point from `src` mode to `dst` mode.
         # TODO: LIDAR
         # only implemented provided Rt matrix in cam-depth conversion
-        if src == Coord3DMode.LIDAR and dst == Coord3DMode.CAM:
+        if (src == Coord3DMode.LIDAR
+                and dst == Coord3DMode.CAM) and rt_mat is None:
             rt_mat = arr.new_tensor([[0, -1, 0], [0, 0, -1], [1, 0, 0]])
-        elif src == Coord3DMode.CAM and dst == Coord3DMode.LIDAR:
+        elif (src == Coord3DMode.CAM
+              and dst == Coord3DMode.LIDAR) and rt_mat is None:
             rt_mat = arr.new_tensor([[0, 0, 1], [-1, 0, 0], [0, -1, 0]])
         elif src == Coord3DMode.DEPTH and dst == Coord3DMode.CAM:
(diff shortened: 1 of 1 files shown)


mmdet3d/core/bbox/structures/coord_3d_mode.py
```diff
@@ -237,11 +239,13 @@ def convert_point(point, src, dst, rt_mat=None):
             else:
                 rt_mat = rt_mat @ rt_mat.new_tensor([[1, 0, 0], [0, 0, 1],
                                                      [0, -1, 0]])
-        elif src == Coord3DMode.LIDAR and dst == Coord3DMode.DEPTH:
+        elif (src == Coord3DMode.LIDAR
```
THU17cyz, line 242: How about keeping the `rt_mat is None` in an inner if branch like the code in `box_3d_mode.py`? This alternative seems easier to understand since the if logic is clearer.

mmdet3d/core/bbox/structures/coord_3d_mode.py
```diff
@@ -220,9 +220,11 @@ def convert_point(point, src, dst, rt_mat=None):
...
-        if src == Coord3DMode.LIDAR and dst == Coord3DMode.CAM:
+        if (src == Coord3DMode.LIDAR
+                and dst == Coord3DMode.CAM) and rt_mat is None:
             rt_mat = arr.new_tensor([[0, -1, 0], [0, 0, -1], [1, 0, 0]])
-        elif src == Coord3DMode.CAM and dst == Coord3DMode.LIDAR:
+        elif (src == Coord3DMode.CAM
+              and dst == Coord3DMode.LIDAR) and rt_mat is None:
             rt_mat = arr.new_tensor([[0, 0, 1], [-1, 0, 0], [0, -1, 0]])
         elif src == Coord3DMode.DEPTH and dst == Coord3DMode.CAM:
             if rt_mat is None:
```
THU17cyz, line 230: Also, maybe it's helpful to add a comment here to point out why LIDAR-CAM conversion is different from DEPTH-CAM conversion? You can say that SUNRGB-D camera calibration files are different from that of KITTI, and currently we keep this hack.

wHao-Wu: Thanks for your advices. I have updated the place of `rt_mat is None` and added the comment to clarify the difference between LIDAR-CAM conversion and DEPTH-CAM conversion.

THU17cyz (approved):

PR #579, opened 2021-05-24
[Enhance] Sampling points based on distance metric

wHao-Wu:
* Sampling points based on distance metric (keeping all the far points and sampling only the near points)
* Rename function `IndoorPointSample` to `PointSample`

### DIFF
--- a/mmdet3d/datasets/pipelines/transforms_3d.py
+++ b/mmdet3d/datasets/pipelines/transforms_3d.py
@@ -1,4 +1,5 @@
 import numpy as np
+import warnings
 from mmcv import is_tuple_of
 from mmcv.utils import build_from_cfg
 
@@ -731,12 +732,13 @@ class PointSample(object):
         Args:
             points (np.ndarray | :obj:`BasePoints`): 3D Points.
             num_samples (int): Number of samples to be sampled.
-            dist_metric (int): The indicator to the near/far boundary.
+            dist_metric (int, optional): Indicator to the near/far boundary.
                 Once given, only the near points will be sampled.
                 Defaults to None.
-            replace (bool): Whether the sample is with or without replacement.
+            replace (bool, optional): Sampling with or without replacement.
                 Defaults to None.
-            return_choices (bool): Whether return choice. Defaults to False.
+            return_choices (bool, optional): Whether return choice.
+                Defaults to False.
 
         Returns:
             tuple[np.ndarray] | np.ndarray:
@@ -748,7 +750,7 @@ class PointSample(object):
             replace = (points.shape[0] < num_samples)
         sample_range = range(len(points))
         if dist_metric:
-            depth = points.tensor[:, 2].numpy()
+            depth = points.coord[:, 2]
             far_inds = np.where(depth > dist_metric)[0]
             near_inds = np.where(depth <= dist_metric)[0]
             if not replace:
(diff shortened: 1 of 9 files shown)


mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -703,22 +703,25 @@ def __repr__(self):
 
 
 @PIPELINES.register_module()
-class IndoorPointSample(object):
```
ZwwWayne, line 706: We should keep the API of `IndoorPointSample` and add deprecated warning in case downstream repos use it in their projects.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -728,8 +731,11 @@ def points_random_sampling(self,
         Args:
             points (np.ndarray | :obj:`BasePoints`): 3D Points.
             num_samples (int): Number of samples to be sampled.
+            dist_metric (int): The indicator to the near/far boundary.
```
ZwwWayne, line 734: (int, optional)

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -757,9 +775,13 @@ def __call__(self, results):
             dict: Results after sampling, 'points', 'pts_instance_mask' \
                 and 'pts_semantic_mask' keys are updated in the result dict.
         """
+        from mmdet3d.core.points import CameraPoints
         points = results['points']
+        if self.dist_metric:
```
ZwwWayne, line 780: Why do we need this constraint? We need to add comment for that.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -740,8 +746,20 @@ def points_random_sampling(self,
         """
         if replace is None:
             replace = (points.shape[0] < num_samples)
-        choices = np.random.choice(
-            points.shape[0], num_samples, replace=replace)
+        sample_range = range(len(points))
+        if dist_metric:
+            depth = points.tensor[:, 2].numpy()
```
ZwwWayne, line 751: We should use the API in the points structure to find the points to be kept. Thus this API can be applied to all coordinates.

wHao-Wu:
@ZwwWayne 

Thanks for your advice. I have updated the code:
* Adding the deprecated warning for IndoorPointSample.
* Adding necessary comments.
* Using the API provided by point structure in `base_points.py`.

wHao-Wu:
I want to add the `IndoorPointSample` into the `__init__` but can not pass the pre-commit. The message in pre-commit always shows:

image

However, there isn't any fixing in my code. So I temporarily reset my code and want more suggestions.

Wuziyi616:
> 
> 
> I want to add the `IndoorPointSample` into the `__init__` but can not pass the pre-commit. The message in pre-commit always shows:
> 
> image
> 
> However, there isn't any fixing in my code after pre-commit. So I temporarily reset my code and want more suggestions.

Maybe that's because yapf and isort has some conflicts here (isort first fix to some style1, and then yapf change it back to original style). You can disable yapf like [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/utils.py#L1-L10) to solve this problem.
```
# yapf: disable
some import code
# yapf: enable
```

wHao-Wu:
> > I want to add the `IndoorPointSample` into the `__init__` but can not pass the pre-commit. The message in pre-commit always shows:
> > image
> > However, there isn't any fixing in my code after pre-commit. So I temporarily reset my code and want more suggestions.
> 
> Maybe that's because yapf and isort has some conflicts here (isort first fix to some style1, and then yapf change it back to original style). You can disable yapf like [here](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/datasets/utils.py#L1-L10) to solve this problem.
> 
> ```
> # yapf: disable
> some import code
> # yapf: enable
> ```

Thanks for your suggestion. After following your instruction, I have passed the pre-commit successfully.

PR #542, opened 2021-05-14
[Feature] Support visualization and browse_dataset on nuScenes Mono-3D dataset

Wuziyi616:
As `NuScenesMonoDataset` inherits from 2D `CocoDataset`, to be coherent with the visualization setting in MMDet3D:

- I add `eval_pipeline` for its config file
- I add some functions to `NuScenesMonoDataset` to support data loading in visualization (e.g. `_extract_data`) following `Custom3DDataset`
- In order to correctly visualize the mono boxes, I add a hacking function to `mmdet3d/core/bbox/structures/utils.py` named `mono_cam_box2vis`

To support Mono-3D in browse_dataset, I let users specify the task of the dataset instead of inferring from config (which is troublesome).

Since we don't have kitti mono config in the current codebase, I didn't test the code on it. But I believe it will work well as KittiMono inherits from NusMono.

PR #583, opened 2021-05-25
[Enhance] Reuse some functions in `Datasets` loading data

Wuziyi616:
See https://github.com/open-mmlab/mmdetection3d/pull/542#discussion_r637499596.

Currently, I only move the staticmethod `self._get_data()` to `utils.py` because the other two functions `self._get_pipeline()` and `self._extract_data()` involve many `self.xxx` calling.

A solution is to use utils functions that take `dataset` as input args, and call `dataset.xxx` as `self.xxx`. But I am not sure if this is a valid coding style? @Tai-Wang can you give me some advice?

### DIFF
--- a/mmdet3d/datasets/utils.py
+++ b/mmdet3d/datasets/utils.py
@@ -1,3 +1,5 @@
+import mmcv
+
 # yapf: disable
 from mmdet3d.datasets.pipelines import (Collect3D, DefaultFormatBundle3D,
                                         LoadAnnotations3D,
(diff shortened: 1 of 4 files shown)


Tai-Wang (approved):

ZwwWayne (approved):

mmdet3d/datasets/utils.py
```diff
@@ -108,3 +110,25 @@ def get_loading_pipeline(pipeline):
         'The data pipeline in your config file must include ' \
         'loading step.'
     return loading_pipeline
+
+
+def extract_result_dict(results, key):
+    """Extract and return the data corresponding to key in result dict.
+
```
ZwwWayne, line 117: Indicate that this module essentially obtain the data from the data container and list or tuple in the description.

Wuziyi616: Add.

Wuziyi616:
@ZwwWayne Merge master. I think it's done.

PR #585, opened 2021-05-25
[Feature] Add `RandomDropPointsColor` transform

Wuziyi616:
A transform function used in 3D segmentation, that set colors of points to all zeros according to a probability. Used in PAConv's paper.

### DIFF
--- a/mmdet3d/datasets/pipelines/transforms_3d.py
+++ b/mmdet3d/datasets/pipelines/transforms_3d.py
@@ -11,6 +11,47 @@ from .data_augment_utils import noise_per_object_v3_
 
 
 @PIPELINES.register_module()
+class RandomDropPointsColor(object):
+    """Randomly set the color of points to all zeros.
+
+    Args:
+        drop_ratio (float): The probability of dropping point colors.
+            Defaults to 0.2.
+    """
+
+    def __init__(self, drop_ratio=0.2):
+        assert isinstance(drop_ratio, (int, float)) and 0 <= drop_ratio <= 1, \
+            f'invalid drop_ratio value {drop_ratio}'
+        self.drop_ratio = drop_ratio
+
+    def __call__(self, input_dict):
+        """Call function to drop point colors.
+
+        Args:
+            input_dict (dict): Result dict from loading pipeline.
+
+        Returns:
+            dict: Results after color dropping, \
+                'points' key is updated in the result dict.
+        """
+        points = input_dict['points']
+        assert points.attribute_dims is not None and \
+            'color' in points.attribute_dims.keys(), \
+            'Expect points have color attribute'
+
+        if np.random.rand() < self.drop_ratio:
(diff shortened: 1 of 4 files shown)


mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -10,6 +10,47 @@
...
+        Args:
+            input_dict (dict): Result dict from loading pipeline.
+
+        Returns:
+            dict: Results after color dropping, \
+                'points' key is updated in the result dict.
+        """
+        points = input_dict['points']
+        assert points.attribute_dims is not None and \
+            'color' in points.attribute_dims.keys(), \
```
THU17cyz, line 39: May omit .keys()

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -10,6 +10,47 @@
...
+            'Expect points have color attribute'
+
+        if np.random.rand() < self.drop_ratio:
+            points.color = points.color * 0.0
+        input_dict['points'] = points
```
THU17cyz, line 44: Is this line necessary?

THU17cyz (approved):

mmdet3d/datasets/pipelines/transforms_3d.py
Wuziyi616: I agree.

mmdet3d/datasets/pipelines/transforms_3d.py
Wuziyi616: Yes not necessary.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -10,6 +10,46 @@
 from .data_augment_utils import noise_per_object_v3_
 
 
+@PIPELINES.register_module()
+class RandomDropPointsColor(object):
+    """Randomly set the color of points to all zeros.
+
```
ZwwWayne, line 16: It seems that this is applied to all the points once the drop is executed. Do we need to support moderately drop some of the points? If it is unnecessary, we may add a description for that in detail, e.g., indicating this is a common practice in xxx.

Wuziyi616: I only see this augmentation in PAConv's paper so I am not sure if it can be called a "common practice".

In my opinion, I think dropping some points' color while keeping others will harm the model's performance. The intra-object color consistency (same objects share similar color, e.g. office chairs are totally black) is an important clue for segmentors. If we only drop some of the points, then the color will be messy and I think it would be bad for the learning process. Dropping all points' color keeps this consistency (though the inter-object difference is destroyed).

I'm inclined to not support partially dropping. We can add this in the future if needed but not now. I will add a link to PAConv's implementation.

PR #586, opened 2021-05-25
[Enhance] Faster implementation of KNN

Wuziyi616:
This new KNN code is borrowed from PAConv. They use heap data structure to accelerate so it's called `knn_heap`.
```
points.shape == [8, 4096, 3]
centers.shape == [8, 64, 3]
k = 32
100 loops:
The current knn needs ~8s, while knn_heap only needs < 0.1s!
```
Should we still keep the current knn in the codebase? After all, they have the same functionality. And knn is not used in our codebase now so this won't cause any BC-breaking. (Or we can do a `import knn_heap as knn` trick in `mmdet3d/ops/__init__.py`?)


Wuziyi616:
My first time working with custom gpu ops. Please point out any mistakes I made.

ZwwWayne:
If the input and output of these two modules are the same, we can simply replace the original one for acceleration. Otherwise, we might need to prevent BC-breaking.

Wuziyi616:
> If the input and output of these two modules are the same, we can simply replace the original one for acceleration. Otherwise, we might need to prevent BC-breaking.

Yes, I modify the wrapper so that the input and output are the same now. Also, `knn` is not used in our current codebase (I understand some users may use it in their projects). I will replace the original one.

THU17cyz (approved):

THU17cyz:
Maybe add a line somewhere telling users the speed difference of two implementations?

Wuziyi616:
> Maybe add a line somewhere telling users the speed difference of two implementations?

I will just replace the old one with the new KNN op since their interface are the same.

PR #598, opened 2021-05-27
[Feature] Support PAConv operation

Wuziyi616:
Implement PAConv layer. PAConv is similar to `Conv2d` or `MLP`, serving as basic building block of a model. The basic idea of PAConv is simple. PAConv dynamically generates convolutional kernels by assembling some weight kernels in its weight bank. Every PAConv has a weight bank which consists of several (typically 16) trainable weight kernels, and a `ScoreNet` which is implemented by MLPs. Given a point pair `p_c is a center` and `p_1 is its neighbor (queried by KNN)`, it uses `ScoreNet` to predict scores for this point pair, and uses the score to assemble weight bank to form one kernel. Then, it runs convolution using the assembled kernel on the point pair and get output features.

**PAConv has two versions of implementation**, one is called `PAConv` and the other one is called `PAConvCUDA`. The differences are:

- `PAConv` is easy to understand, and can run on both CPU as well as GPU. Its inputs are features already grouped by KNN, so it only needs to compute scores and assemble weights and get the output.
- `PAConvCUDA` is tricky, and can only run on GPU because it utilizes a custom cuda op `assign_score_withk`. Its inputs are features **NOT** grouped by KNN, and queries KNN on the fly (hard to describe). This saves memory because we don't need to pre-compute KNN features which have very large tensor size (typically `(B, C, npoint, K)`).

The operation is somewhat complex and I am not sure if I should write its detailed process in the comment. Also, the original code is dirty, although I tried my best to add comments and made it clear, it may still not meet our standard. Please kindly help me improve it (e.g. some function's name).


Wuziyi616:
Would be better if you read the paper before reviewing my code.

Wuziyi616 (commented):
I think this PR is ready for review @Tai-Wang @ZwwWayne

mmdet3d/ops/paconv/paconv.py
```diff
@@ -0,0 +1,350 @@
...
+    Args:
+        in_channels (int): Input channels of point features.
+        out_channels (int): Output channels of point features.
+        num_kernels (int): Number of weight kernels in the weight bank.
+        norm_cfg (dict, optional): Type of normalization method.
+            Defaults to dict(type='BN2d', momentum=0.1).
+        act_cfg (dict, optional): Type of activation method.
+            Defaults to dict(type='ReLU', inplace=True).
+        scorenet_input (str, optional): Type of input to ScoreNet.
+            Can be 'identity', 'neighbor' or 'ed7'. Defaults to 'ed7'.
```
Tai-Wang, line 115: ed7 is quite strange, and 'neighbor' actually refers to 'identity'+'neighbor'?

mmdet3d/ops/paconv/paconv.py
```diff
@@ -0,0 +1,350 @@
...
+        # where C can be in_c or (2 * in_c)
+        if weight_bank_init == 'kaiming':
+            weight_init = nn.init.kaiming_normal_
+        elif weight_bank_init == 'xavier':
+            weight_init = nn.init.xavier_normal_
+        else:
+            raise NotImplementedError(
+                f'unsupported weight bank init method {weight_bank_init}')
+
+        self.m = num_kernels
```
Tai-Wang, line 186: `m` is so abstract? still use self.num_kernels would be better?

mmdet3d/ops/paconv/paconv.py
```diff
@@ -0,0 +1,350 @@
...
+        scorenet_cfg_['mlp_channels'].append(self.m)
+        self.scorenet = ScoreNet(**scorenet_cfg_)
+
+        self.bn = build_norm_layer(norm_cfg, out_channels)[1] if \
+            norm_cfg is not None else None
+        self.activate = build_activation_layer(act_cfg) if \
+            act_cfg is not None else None
+
+    def forward(self, points_xyz: torch.Tensor,
+                features: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
Tai-Wang, line 205: Actually we do not need such specific types for these input params because we have explanations in the docstring. Some previous codes keep these (due to some legacy issues) but we can just simplify them in our new codes.
Just write as forward(self, points_xyz, features) to be more concise.

mmdet3d/ops/paconv/paconv.py
```diff
@@ -0,0 +1,350 @@
...
+
+        # pre-compute features for points and centers separately
+        # features is [B, in_c, N], weight_bank is [C, m * out_dim]
+        point_feat, center_feat = assign_kernel_withoutk(
+            features, self.weight_bank, self.m)
+
+        # aggregate features using custom cuda op
+        new_features = assign_score_cuda(
+            scores, point_feat, center_feat, points_idx,
+            'sum').contiguous()  # [B, out_c, npoint, K]
```
Tai-Wang, line 343: Can we merge the general version and CUDA-only version? The main differences, in my opinion, are the additional input of `point_idx` and the way to assign scores/aggregate features?

mmdet3d/ops/paconv/utils.py
```diff
@@ -0,0 +1,85 @@
+import torch
+
+
+def calc_euclidian_dist(xyz1, xyz2):
+    """Calculate the Euclidian distance between two groups of points.
+
+    Args:
+        xyz1 (torch.Tensor): (N, 3)
+        xyz2 (torch.Tensor): (N, 3)
```
Tai-Wang, line 9: Refine the docstrings

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
```diff
@@ -0,0 +1,214 @@
...
+      fprintf(stderr, "CUDA kernel failed : %s\n%s at L:%d in %s\n",  \
+              cudaGetErrorString(err), __PRETTY_FUNCTION__, __LINE__, \
+              __FILE__);                                              \
+      exit(-1);                                                       \
+    }                                                                 \
+  } while (0)
+
+
+// input: points(B,N0,M,O), centers(B,N0,M,O), scores(B,N1,K,M), knn_idx(B,N1,K)
+// ouput: fout(B,O,N)
```
Tai-Wang, line 39: output

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
```diff
@@ -0,0 +1,214 @@
...
+// algo: fout(b,i,k,j) = s(b,i,k,m)*p(b,c(i),k,m,j) =  s(b,i,k,m)*p(b,i(k),m,j)
+//       i(k) = idx(b,i,k)
+//      sum: fout(b,i,j) = fout(b,i,j) + s(b,i,k,m)*p(b,i,k,m,j)
+//      avg: fout(b,i,j) = sum(fout(b,i,k,j)) / k
+//      max: fout(b,i,j) = max(fout(b,i,k,j), sum(s(b,i,k,m)*p(b,i,k,m,j)))
+
+
+__global__ void assign_score_withk_forward_kernel(const int B, const int N0, const int N1,
+                                                    const int M, const int K, const int O, const int aggregate,
```
Tai-Wang, line 48: adjust the indents

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
```diff
@@ -0,0 +1,214 @@
...
+        assert (n < N1);
+        atomicAdd(output + b*N1*O*K + o*N1*K + n*K + k,
+            points[b*N0*M*O + kn*M*O + m*O + o] * scores[b*N1*K*M + n*K*M + k*M + m]
+                - centers[b*N0*M*O + cn*M*O + m*O + o] * scores[b*N1*K*M + n*K*M + k*M + m]);
+    }
+}
+
+
+__global__ void assign_score_withk_backward_points_kernel(const int B, const int N0, const int N, const int M,
+                                                            const int K, const int O, const int aggregate,
```
Tai-Wang, line 82: indents

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
```diff
@@ -0,0 +1,214 @@
...
+            int cn = knn_idx[b*N*K + n*K + 0];
+            if (kn >= N0 || kn < 0) { // if index overflows, it is out of the neighborhood range
+                continue;
+            }
+            atomicAdd(grad_points + b*N0*M*O + kn*M*O + m*O + o,
+                scores[b*N*K*M + n*K*M + k*M + m] * grad_out[b*O*N*K + o*N*K + n*K + k]);
+            atomicAdd(grad_centers + b*N0*M*O + cn*M*O + m*O + o,
+                - scores[b*N*K*M + n*K*M + k*M + m] * grad_out[b*O*N*K + o*N*K + n*K + k]);
+            //grad_points[b*N*M*O + kn*M*O + m*O + o] += 2 * scores[b*N*K*M + n*K*M + k*M + m] * grad_out[b*O*N + o*N + n];
+            //grad_points[b*N*M*O + n*M*O + m*O + o] -= scores[b*N*K*M + n*K*M + k*M + m] * grad_out[b*O*N + o*N + n];
```
Tai-Wang, line 109: Clean these unused comments? Out of curiosity, the 2 in the first equation corresponds to sth. in the line 104-105 or just a typo?

mmdet3d/ops/paconv/paconv.py
Wuziyi616: Yes, I have changed this variable name in later [branch](https://github.com/Wuziyi616/mmdetection3d/blob/paconv_model/mmdet3d/ops/paconv/paconv.py#L187). Maybe we can just leave it as is and change in later merge of that branch?

mmdet3d/ops/paconv/paconv.py
Wuziyi616: Fix PAConv and PAConvCUDA's forward func.

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
Wuziyi616: Fix.

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
Wuziyi616: Fix.

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
Wuziyi616: Fix all indents in this cu file.

mmdet3d/ops/paconv/src/assign_score_withk_cuda.cu
Wuziyi616: I think they are just typos.

mmdet3d/ops/paconv/utils.py
Wuziyi616: Done.

mmdet3d/ops/paconv/paconv.py
Wuziyi616: Changed.

mmdet3d/ops/paconv/paconv.py
Wuziyi616: Refactor PAConv to share more functions.

Tai-Wang (approved):

Issue #347, opened 2021-03-09
Bugs of BackgroundPointsFilter 


meng-zha:
[BackgroundPointsFilter](https://github.com/open-mmlab/mmdetection3d/blob/7f9b186e82d7945cb3bfe6fb4bc110dcb1d570ae/mmdet3d/datasets/pipelines/transforms_3d.py#L676) transformer changes the LiDARInstance3DBox's tensor to the gravity center while it should be the bottom_center. The _box_np_ops.points_in_rbbox_ deals with the boxes of bottom center as well. So there may be some bugs here.

meng-zha:
@encore-zhou

PR #609, opened 2021-06-01
[Fix] Fix bg point filter 

xiliu8006:
fix the issue(#347)


configs/3dssd/3dssd_kitti-3d-car.py
```diff
@@ -87,8 +87,8 @@
 ]
 
 data = dict(
-    samples_per_gpu=4,
-    workers_per_gpu=4,
+    samples_per_gpu=2,
```
Wuziyi616, line 90: This config file seems strange to me. It doesn't have `batch_size x num_GPUs` in its name (such as `2x8`)?

Wuziyi616 (commented):
The issue said this [line](https://github.com/open-mmlab/mmdetection3d/blob/7f9b186e82d7945cb3bfe6fb4bc110dcb1d570ae/mmdet3d/models/dense_heads/ssd_3d_head.py#L426) is also strange. Is it also caused by this bug?

configs/3dssd/3dssd_kitti-3d-car.py
xiliu8006: I forget to modify it, sorry

Wuziyi616: What about the filename? Should we change it to something like `3dssd_**4x8**_kitti-3d-car.py`?

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -1127,11 +1127,11 @@ def __call__(self, input_dict):
         points = input_dict['points']
         gt_bboxes_3d = input_dict['gt_bboxes_3d']
 
-        gt_bboxes_3d_np = gt_bboxes_3d.tensor.numpy()
-        gt_bboxes_3d_np[:, :3] = gt_bboxes_3d.gravity_center.numpy()
+        gt_bboxes_3d_np = gt_bboxes_3d.tensor.clone().numpy()
```
ZwwWayne, line 1130: Add comments indicating why clone is necessary here.

ZwwWayne (approved):

PR #618, opened 2021-06-04
[Feature] Support PAConv model training

Wuziyi616:
Will benchmark in this PR. Stay tuning.


configs/_base_/models/paconv_ssg.py
```diff
@@ -0,0 +1,46 @@
...
+        conv_cfg=dict(type='Conv1d'),
+        norm_cfg=dict(type='BN1d'),
+        act_cfg=dict(type='ReLU'),
+        loss_decode=dict(
+            type='CrossEntropyLoss',
+            use_sigmoid=False,
+            class_weight=None,  # should be modified with dataset
+            loss_weight=1.0)),
+    # regularization loss for PAConv's weight kernels
+    regularization_loss=dict(
```
ZwwWayne, line 42: We usually use `loss_regularization`

configs/paconv/paconv_cuda_ssg_8x2_step_100e_s3dis_seg-3d-13class.py
```diff
@@ -0,0 +1,75 @@
...
+    dict(
+        type='IndoorPatchPointSample',
+        num_points=num_points,
+        block_size=1.0,
+        sample_rate=1.0,
+        ignore_index=len(class_names),
+        use_normalized_coord=True),
+    dict(
+        type='GlobalRotScaleTrans',
+        rot_range=[-3.141592653589793, 3.141592653589793],
```
ZwwWayne, line 37: may indicate how we obtain this number, e.g., it is [-pi, pi]

mmdet3d/models/losses/paconv_correlation_loss.py
```diff
@@ -0,0 +1,108 @@
...
+    elif reduction == 'none':
+        pass
+    else:
+        raise NotImplementedError
+
+    return corr_loss
+
+
+@LOSSES.register_module()
+class PAConvCorrelationLoss(nn.Module):
```
ZwwWayne, line 75: if we use PAConvCorrelationLoss here, we should consider change `loss_regularization` to `loss_correlation`.

mmdet3d/models/losses/paconv_correlation_loss.py
```diff
@@ -0,0 +1,108 @@
...
+    """Calculate correlation loss of weight kernels in PAConv's weight bank.
```
ZwwWayne, line 76: of kernel weights in PAConv

mmdet3d/models/losses/paconv_correlation_loss.py
```diff
@@ -0,0 +1,108 @@
...
+    # take upper triangular part excluding diagonal since we only compute
+    # correlation between different kernels once
+    # the square is to ensure positive loss, refer to:
+    # https://github.com/CVMI-Lab/PAConv/blob/main/scene_seg/tool/train.py#L208
+    corr = torch.sum(torch.triu(cosine_sims, diagonal=1)**2)
+
+    return corr
+
+
+def paconv_correlation_loss(modules, reduction):
```
ZwwWayne, line 44: can we directly use the weight_reduce_loss wrapper in MMDetection? https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/losses/utils.py#L28

mmdet3d/models/segmentors/base.py
```diff
@@ -16,6 +16,12 @@ class Base3DSegmentor(BaseSegmentor):
     data_dict and use a 3D seg specific visualization function.
     """
 
+    @property
+    def with_regularization_loss(self):
```
ZwwWayne, line 20: with_correlation_loss

mmdet3d/models/segmentors/base.py
```diff
@@ -63,18 +69,18 @@ def forward(self, return_loss=True, **kwargs):
     def show_results(self,
                      data,
                      result,
-                     palette=None,
```
ZwwWayne, line 66: Will change the order of argument cause BC-breaking?

mmdet3d/models/segmentors/encoder_decoder.py
```diff
@@ -130,6 +142,19 @@ def _auxiliary_head_forward_train(self, x, img_metas, pts_semantic_mask):
 
         return losses
 
+    def _regularization_loss_forward_train(self):
+        """Calculate regularization loss for model weight in training."""
+        losses = dict()
+        if isinstance(self.regularization_loss, nn.ModuleList):
+            for idx, reg_loss in enumerate(self.regularization_loss):
```
ZwwWayne, line 149: reg_loss will cause ambiguity as we have `reg_loss` for regression.

configs/_base_/models/paconv_cuda_ssg.py
```diff
@@ -0,0 +1,7 @@
+_base_ = './paconv_ssg.py'
+
+model = dict(
+    backbone=dict(
+        sa_cfg=dict(
+            type='PAConvSAModuleCUDA',
```
ZwwWayne, line 6: CUDAPAConvSAModule might be better

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -2,22 +2,21 @@
...
+from mmdet3d.ops import (GroupAll, PAConv, Points_Sampler, QueryAndGroup,
+                         gather_points)
 from .builder import SA_MODULES
 
 
-@SA_MODULES.register_module()
-class PointSAModuleMSG(nn.Module):
-    """Point set abstraction module with multi-scale grouping used in
-    Pointnets.
+class _PointSAModuleBase(nn.Module):
```
ZwwWayne, line 12: BasePointSAModule

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -144,44 +185,116 @@ def forward(
...
+                new_features = new_features[0]
+
+            # (B, mlp[-1], num_point)
+            new_features = self._pool_features(new_features)
             new_features_list.append(new_features)
 
         return new_xyz, torch.cat(new_features_list, dim=1), indices
 
 
+@SA_MODULES.register_module()
```
ZwwWayne, line 216: we can use `@SA_MODULES.register_module(name='PointSAMSG')` to simplify the name

mmdet3d/models/segmentors/base.py
Wuziyi616: I think it won't. Because this function is only called inside the codebase of MMDet3D (at least we don't have instructions that require calling this function).

configs/_base_/models/paconv_cuda_ssg.py
Wuziyi616: Do you think `PAConvCUDASAModule` is better? Since we use `PAConv` and `PAConvCUDA` for these two basic conv operation.

mmdet3d/ops/pointnet_modules/point_sa_module.py
Wuziyi616: Fix.

mmdet3d/ops/pointnet_modules/point_sa_module.py
Wuziyi616: This will introduce a BC-breaking as some other models also use SA modules. Do you think we need to do this?

mmdet3d/models/losses/paconv_correlation_loss.py
Wuziyi616: I agree that we may need to change the name of `loss_regularization`. But don't you think `loss_correlation` is a little bit ad-hoc. I don't think any other paper will use a "correlation" loss. What about changing `PAConvCorrelationLoss` to `PAConvRegularizationLoss` so that we can retain the interface `loss_regularization`, which may be used by other segmentors in the future?

mmdet3d/models/segmentors/base.py
Wuziyi616: Still, I think correlation loss may not be general enough?

configs/_base_/models/paconv_cuda_ssg.py
Wuziyi616: Change to `PAConvCUDASAModule`.

configs/paconv/paconv_cuda_ssg_8x2_step_100e_s3dis_seg-3d-13class.py
Wuziyi616: Add.

configs/_base_/models/paconv_ssg.py
Wuziyi616: Fix. Also all the following parts.

mmdet3d/models/segmentors/encoder_decoder.py
Wuziyi616: Fix to regularize_loss.

mmdet3d/models/losses/paconv_correlation_loss.py
Wuziyi616: Fix all weight kernels to kernel weights.

mmdet3d/models/losses/paconv_correlation_loss.py
Wuziyi616: rename to `PAConvRegularizationLoss`

mmdet3d/models/losses/paconv_correlation_loss.py
Wuziyi616: Use it to perform loss reduction.

mmdet3d/core/visualizer/show_result.py
```diff
@@ -131,7 +131,7 @@ def show_seg_result(points,
                     filename,
                     palette,
                     ignore_index=None,
-                    show=False,
```
ZwwWayne, line 134: Why this PR modify this?

ZwwWayne:
Please @Tai-Wang have a look. We can start to prepare the model checkpoints for release.

mmdet3d/models/decode_heads/paconv_head.py
```diff
@@ -0,0 +1,63 @@
...
+        # so we need to rebuild it here
+        delattr(self, 'pre_seg_conv')
+        self.pre_seg_conv = ConvModule(
+            fp_channels[-1][-1],
+            self.channels,
+            kernel_size=1,
+            bias=False,
+            conv_cfg=self.conv_cfg,
+            norm_cfg=self.norm_cfg,
+            act_cfg=self.act_cfg)
```
Tai-Wang, line 38: Direct assigning a new module to pre_seg_conv can not update it? I am curious about the necessity of delattr here?

mmdet3d/models/losses/paconv_regularization_loss.py
```diff
@@ -0,0 +1,106 @@
...
+    regularization loss. The cosine similarity is used as metrics.
+
+    Args:
+        conv (nn.Module): A Conv modules to be regularized.
+            Currently we only support `PAConv` and `PAConvCUDA`.
+
+    Returns:
+        torch.Tensor: Correlations between each kernel weights in weight bank.
+    """
+    assert isinstance(conv, (PAConv, PAConvCUDA))
```
Tai-Wang, line 20: Only support PAConv? Add the corresponding assertion message

Tai-Wang (commented):
Almost ready for merging except some details needed to be fixed

mmdet3d/models/losses/paconv_regularization_loss.py
Wuziyi616: No. This loss regularizes the weight kernels in both `PAConv` and `PAConvCUDA`. Basically, `PAConv` and `PAConvCUDA` are just two different implementations with almost the same in-out. `PAConvCUDA` uses custom CUDA op to accelerate computation and reduce GPU memory usage.

I will update the assertion info.

mmdet3d/models/decode_heads/paconv_head.py
Wuziyi616: Oh I yes you are right. There is nothing special here. I will remove that `delattr` line.

mmdet3d/core/visualizer/show_result.py
Wuziyi616: Oh, that's because when I benchmark the model, I want to do some visualization to help me decide if the output is correct. Then I realized the previous code has this bug so I fix it in this PR. The order here will cause error when setting `--show` in `test.py` args.

Wuziyi616: I just want to make it consistent with `show_result()`. In my opinion, `show_result()`, `show_seg_result()` and `show_multi_modality_result()` are three similar visualization functions in the same file, we should keep their default args consistent. As `show_result()` has `True` for default value of `show`, I am changing the other two to `True`, too.

I have checked the codebase and this won't cause any BC-breaking. `show` args is always assigned when calling this function.

mmdet3d/models/decode_heads/paconv_head.py
Wuziyi616: I agree that this shouldn't be modified in this PR. I am changing it back. Maybe do these modifications in a new PR.

mmdet3d/core/visualizer/show_result.py
Wuziyi616: I agree that this shouldn't be modified in this PR. I am changing it back. Maybe do these modifications in a new PR.

Wuziyi616: Changed them in #783

mmdet3d/models/decode_heads/paconv_head.py
Wuziyi616: Changed them in #783

mmdet3d/models/losses/paconv_regularization_loss.py
Wuziyi616: Changed them in #783

PR #621, opened 2021-06-07
[Enhance] Confirm the input does not be changed in bg_point_filter unit test

xiliu8006:
add unittest to confirm the input is not be modified


Wuziyi616:
You forgot to remove the `swp` file.

Wuziyi616 (approved):

tests/test_data/test_pipelines/test_augmentations/test_transforms_3d.py
```diff
@@ -623,6 +624,8 @@ def test_background_points_filter():
     assert repr_str == expected_repr_str
     assert points.shape == (800, 4)
     assert np.allclose(orig_points, points)
+    assert np.allclose(input_dict['gt_bboxes_3d'].tensor.numpy(),
```
ZwwWayne, line 627: Should we use np.equal rather than np.allclose? so as line 626.

ZwwWayne (approved):

PR #652, opened 2021-06-16
[Enhance] Support PointNet2 MSG w/ FP

wHao-Wu:
* Support PointNet2 w/ FP Module
* Set the default value of fp_channels to an empty tuple in order to be compatible with previous projects.


Wuziyi616:
This modification is reasonable. However, I think these FP modules should not be in SA modules. Because I think they are more like `Neck` so shouldn't be placed in `Backbone`. Maybe do this in a later refactor. @ZwwWayne what do you think?

Wuziyi616 (approved):

Issue #620, opened 2021-06-05
ImVoxelNet


filaPro:
Hi, @Tai-Wang and mmdetection3d team,

This issue continues the discussion in https://github.com/saic-vul/imvoxelnet/issues/7. If you are sure, it would be great to have ImVoxelNet in this codebase, I can make a PR. I would like to clarify some points first.

1. SUN RGB-D. Unfortunately there are 2 monocular benchmarks: [PerspectiveNet](https://arxiv.org/abs/1912.07744) with 30 classes, [Total3DUnderstanding](https://github.com/yinyunie/Total3DUnderstanding) with 37 classes + camera pose estimation + 3d room layout estimation. In our paper we also present metrics for your VoteNet benchmark with 10 classes. Do you think this VoteNet option is enough for the beginning? One more question here. For SUN RGB-D we use differentiable rotated IoU from [here](https://github.com/lilanxiao/Rotated_IoU). Do you want to adapt in to mmdetection3d?

2. ScanNet. Current ScanNetDataset doesn't deal with RGB images. Do we need separate MultiViewScanNet dataset and MultiViewPipeline handling random images sampling on training? Also dataset preparation instruction need to be updated.

3. KITTI. No questions here.

4. nuScenes. Following [OFTNet](https://arxiv.org/abs/1811.08188) and [MonoDIS](https://ieeexplore.ieee.org/abstract/document/9200697) we compare only on `car` category. I think it not enough interesting for community and we don't need to make separate benchmark here.

Tai-Wang:
Hi @filaPro ,

Thanks again for your willingness to contribute. For these questions, in my opinion:

1. I prefer to beginning with our 10-class setting, which can provide an intuitive comparison with methods based on other modalities. Also it would be easier and clearer in terms of the data preparation.
2. If we need new preprocessing scripts and loading pipelines, maybe you can refer to how we add image processing to KITTI and nuScenes. I guess new data converter functions and classes are similarly needed.
4. Yes, I agree with you. Maybe we can try to extend your method to multi-class detection in the future.

Generally speaking, I would recommend you split the modules into several small PRs such that there will be not much burden for us to review and discuss each one. You can also refer to how I push the FCOS3D to this repo. For others, you can basically follow your idea because you are definitely more familiar with your codes, and we can discuss them after PRs are created.

filaPro:
Hi @Tai-Wang ,

[...]

PR #653, opened 2021-06-16
[Fix] Unify camera poses

filaPro:
Replace `calib['K']` and `calib['Rt']` with `img_meta['depth2img']` for `SUN RGB-D` to be like `img_meta['lidar2img']` for outdoor datasets. More discussion can be found in [#620#issuecomment-859664418](https://github.com/open-mmlab/mmdetection3d/issues/620#issuecomment-859664418).

This PR affects visualization on `SUN RGB-D` and `ImVoteNet` model. I tested these 2 things. Validation `mAP` for `SUN RGB-D` improved from `64.04` to `64.61`.


filaPro:
Sorry, will think about failed tests later.

configs/imvotenet/imvotenet_stage2_16x8_sunrgbd-3d-10class.py
```diff
@@ -230,7 +230,7 @@
                 type='DefaultFormatBundle3D',
                 class_names=class_names,
                 with_label=False),
-            dict(type='Collect3D', keys=['img', 'points', 'calib'])
```
Wuziyi616, line 233: Please also remove `'calib'` key from `Collect3D` in `eval_pipeline`.

mmdet3d/apis/inference.py
```diff
@@ -155,23 +155,25 @@ def inference_multi_modality_detector(model, pcd, image, ann_file):
...
         seg_fields=[])
-
-    # depth map points to image conversion
-    if box_mode_3d == Box3DMode.DEPTH:
-        data.update(dict(calib=info['calib']))
-
     data = test_pipeline(data)
 
+    # TODO: this code is dangerous and dataset-specific. Move lidar2img and
+    #       depth2img to .pkl annotations in the future.
```
Wuziyi616, line 161: I am a little confused why you think this is dangerous? Are there any cases where `lidar2img` or `depth2img` is not calculated like this?

mmdet3d/core/bbox/structures/utils.py
```diff
@@ -111,12 +111,13 @@ def get_box_type(box_type):
...
 
-def points_cam2img(points_3d, proj_mat):
+def points_cam2img(points_3d, proj_mat, return_z=False):
     """Project points from camera coordicates to image coordinates.
 
     Args:
-        points_3d (torch.Tensor): Points in shape (N, 3)
+        points_3d (torch.Tensor): Points in shape (N, 3).
         proj_mat (torch.Tensor): Transformation matrix between coordinates.
+        return_z (bool): Return third dimension if True.
```
Wuziyi616, line 120: `return_z (bool, optional): Return third dimension if True. Defaults to False.`

mmdet3d/core/visualizer/image_vis.py
```diff
@@ -120,6 +120,7 @@ def draw_lidar_bbox3d_on_img(bboxes3d,
     return plot_rect3d_on_img(img, num_bbox, imgfov_pts_2d, color, thickness)
 
 
+# TODO: remove third parameter in all functions here in favour of img_metas
```
Wuziyi616, line 123: We keep this param because we want to keep the interface of all three projection functions the same, so that we can simply call it [here](https://github.com/open-mmlab/mmdetection3d/blob/c1f6bbab5582957d8ddfb487f9df02e81560e93e/mmdet3d/core/visualizer/show_result.py#L232).

Wuziyi616 (commented):
Good job! Overall the code is much cleaner and easy to follow now. Can you kindly post a few visualization of ImVoteNet so that we can make sure the correctness of this PR? (Also some `browse_dataset` results on SUN RGB-D multi-modality task if you have time)

mmdet3d/core/visualizer/image_vis.py
filaPro: Yes I saw this reason, but other functions in this file use either `img_metas` or this third parameter. So we can probably drop one of them. I prefer to keep `img_metas`, with `lidar2img`, `depth2img` or `cam2img`. Am I missing something?

mmdet3d/apis/inference.py
filaPro: The thing is that all this function is dataset-specific, however it has a general name `inference_multi_modality_detector`. Such things as `R0_rect`, `Tr_velo_to_cam` are KITTI-specific. Others like `info['calib']['Rt']` are SUN RGN-D-specific. What if we have for example nuScenes multi-modality detector? I think if we want to keep this API, it can be refactored in the future. For example by putting `lidar2img` and `depth2img` to `.pkl` annotations preliminarily. For sure can remove this `todo`, if you disagree.

mmdet3d/core/visualizer/image_vis.py
Wuziyi616: Ah, I got it. I agree with you. We can do this in this PR?

mmdet3d/apis/inference.py
Wuziyi616: I agree with you. Let's keep this `TODO`.

filaPro:
Also the CI is broken. Probably not by my commit.

mmdet3d/core/visualizer/image_vis.py
filaPro: This PR affects only SUN RGB-D. If you do not mind I will refactor KITTI matrices and this function in the next PR.

Wuziyi616: Sure. Thank you for your effort!

Wuziyi616:
> 
> 
> Also the CI is broken. Probably not by my commit.

Yes it's because of MMSeg version. Will raise a PR to fix it.

filaPro:
Hi @Wuziyi616 ,
I checked `browse_dataset` as you requested and fixed `calib` there. Now `ImVoteNet` visualization and `browse_dataset` visualization are the same for `master` and current branch. The results are in https://disk.yandex.ru/d/d_22eIbR_2yJyA .

Tai-Wang (approved):

mmdet3d/core/bbox/structures/utils.py
```diff
@@ -111,12 +111,14 @@ def get_box_type(box_type):
     return box_type_3d, box_mode_3d
 
 
-def points_cam2img(points_3d, proj_mat):
+def points_cam2img(points_3d, proj_mat, return_z=False):
```
THU17cyz, line 114: How about keeping the same with https://github.com/open-mmlab/mmdetection3d/blob/a6ee9335682967ad2d697f80cb79991028bb966b/mmdet3d/core/bbox/box_np_ops.py#L338

Wuziyi616:
Hi @filaPro, please merge master so that the CI failure caused by MMSeg version incompatibility will be solved. Then we can run CI to check your PR's correctness. Thanks.

filaPro:
@Wuziyi616 Done.
@THU17cyz Fixed your comments.

Wuziyi616 (approved):
Great work!

THU17cyz (approved):

ZwwWayne (approved):

PR #656, opened 2021-06-17
[Fix] Fix MMSeg max version

Wuziyi616:
MMDet3D is currently compatible with mmseg>=0.13.0, remove the max version limit.

Modify installation instruction to install mmdet 2.11.0.

**Edit**: we need to fix mmseg to 0.14.0 because of [this](https://github.com/open-mmlab/mmsegmentation/commit/db44d16e023fd248280943c0bac79b9b5ce11042) commit.

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -99,7 +99,7 @@ Note:
 MMDetection3D v0.14.0 is only compatible with MMDetection version `mmdet>=2.10.0, <=2.11.0`. The future versions will only support `mmdet>=2.12.0` since v0.15.0 (to be released in July).
 
 ```shell
-pip install git+https://github.com/open-mmlab/mmdetection.git
+pip install mmdet==2.11.0
 ```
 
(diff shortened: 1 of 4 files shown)


wHao-Wu:
@Wuziyi616 
Is it necessary to update the mmseg version in the dockerfile?

docs/getting_started.md
```diff
@@ -99,29 +99,31 @@ Note:
...
-pip install git+https://github.com/open-mmlab/mmdetection.git
+pip install mmdet==2.11.0
 ```
 
 Optionally, you could also build MMDetection from source in case you want to modify the code:
 
 ```shell
 git clone https://github.com/open-mmlab/mmdetection.git
 cd mmdetection
+git checkout 2894516  # switch to v2.11.0 branch
```
Tai-Wang, line 110: git checkout v2.11.0 should be clearer?

Wuziyi616: Yes.

PR #664, opened 2021-06-21
Scannet doc

hjin2902:
Add docs of ScanNet including:

- data preparation: explanation about raw data and annotations
- train pipeline: load data and pre-processing
- metrics: evaluation

### DIFF
--- a/docs/datasets/scannet.md
+++ b/docs/datasets/scannet.md
@@ -40,11 +40,9 @@ Under folder `scans` there are overall 1201 train and 312 validation folders in 
 
 Export ScanNet data by running `python batch_load_scannet_data.py`. The main steps include:
 
-- Load label map file
-- Load raw point cloud data
-- Load semantic and instance label for each point
-- Generate 3d bounding box for each scan
-- Save all raw data and annotations
+- Export original files to point cloud, instance label, semantic label and bounding box file.
+- Downsample raw point cloud and filter invalid classes.
+- Save point cloud data and relevant annotation files.
 
  And the core function `export` in `load_scannet_data.py` is as follows:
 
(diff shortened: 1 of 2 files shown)


docs/datasets/scannet.md
```diff
@@ -0,0 +1,403 @@
+# ScanNet
+
+## Dataset preparation
+
+For the overall process, please refer to [scannet](https://github.com/open-mmlab/mmdetection3d/blob/master/data/scannet/README.md/).
```
Tai-Wang, line 5: refer to the \[README\]\(https://github.com/open-mmlab/mmdetection3d/blob/master/data/scannet/README.md/) page for ScanNet

docs/datasets/scannet.md
```diff
@@ -0,0 +1,403 @@
...
+
+### Export ScanNet data
+
```
Tai-Wang, line 8: Add a sentence to summarize what we do by exporting ScanNet data

Tai-Wang (commented):
Maybe we can simplify the code scripts and replace them with your summary for the data processing steps. Too many codes will make the doc unnecessarily too long.

docs/datasets/scannet.md
```diff
@@ -0,0 +1,351 @@
...
+    return mesh_vertices, label_ids, instance_ids, unaligned_bboxes, \
+        aligned_bboxes, object_id_to_label_id, axis_align_matrix
+
+```
+
+After exporting each scan, the raw point cloud could be downsampled, e.g. to 50000. In addition, invalid semantic label outside of `nyu40id` standard should be filtered. Finally, the point cloud data, semantic labels, instance labels and ground truth bounding boxes should be saved in `.npy` files. The overall `export_one_scan` procedure is as follows:
+
+```python
+def export_one_scan(scan_name,
```
ZwwWayne, line 143: can be simplified.

ZwwWayne (approved):

PR #667, opened 2021-06-22
[Enhance] Sampling points based on distance metric

wHao-Wu:
* Sample points based on distance metric (keeping all the far points and sampling only the near points).
* Rename function IndoorPointSample to PointSample (with backward compatibility).
* Support `new_points = points[4:12, 2]` with single attribute in which points are `BasePoints` object.


mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -890,6 +911,43 @@ def __call__(self, results):
...
+        num_points (int): Number of points to be sampled.
+    """
+
+    def __init__(self, num_points):
+        warnings.warn(
+            'IndoorPointSample is deprecated in favor of PointSample')
+        super(IndoorPointSample, self).__init__(num_points)
+        self.num_points = num_points
+
+    def __call__(self, results):
```
ZwwWayne, line 940: We do not need to override __call__.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -890,6 +911,43 @@ def __call__(self, results):
...
+    """Indoor point sample.
+
+    Sampling data to a certain number.
+    NOTE: IndoorPointSample is deprecated in favor of PointSample
+
+    Args:
+        num_points (int): Number of points to be sampled.
+    """
+
+    def __init__(self, num_points):
```
ZwwWayne, line 934: should use `__init__(self, *args, **kwargs):` and directly pass it to the super class.

mmdet3d/core/points/base_points.py
```diff
@@ -303,6 +305,10 @@ def __getitem__(self, item):
                 item = list(item)
                 item[1] = list(range(start, stop, step))
                 item = tuple(item)
+            elif isinstance(item[1], int):
```
ZwwWayne, line 308: Should add unit tests for this.

ZwwWayne:
Also update configs.

wHao-Wu:
@ZwwWayne 
Whether we need to update the BC-breaking doc [here](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/compatibility.md)?

ZwwWayne:
> @ZwwWayne
> Whether we need to update the BC-breaking doc [here](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/compatibility.md)?

We should.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -890,6 +911,31 @@ def __call__(self, results):
...
+    Args:
+        num_points (int): Number of points to be sampled.
+    """
+
+    def __init__(self, *args, **kwargs):
+        warnings.warn(
+            'IndoorPointSample is deprecated in favor of PointSample')
+        super(IndoorPointSample, self).__init__(*args, **kwargs)
+
     def __repr__(self):
```
ZwwWayne, line 939: Do we need to keep this function?

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,61 @@ def __repr__(self):
...
         Sample points to a certain number.
 
         Args:
             points (np.ndarray | :obj:`BasePoints`): 3D Points.
             num_samples (int): Number of samples to be sampled.
-            replace (bool): Whether the sample is with or without replacement.
-            Defaults to None.
-            return_choices (bool): Whether return choice. Defaults to False.
-
+            dist_metric (int, optional): Indicator to the near/far boundary.
```
ZwwWayne, line 848: sample_range might be more clear, indicating the range where the points will be sampled.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,61 @@ def __repr__(self):
...
+        if sample_range:
+            depth = points.coord[:, 2]
+            far_inds = np.where(depth > sample_range)[0]
+            near_inds = np.where(depth <= sample_range)[0]
+            if not replace:
+                # Only sampling the near points when len(points) >= num_samples
+                point_range = near_inds
+                num_samples -= len(far_inds)
+        choices = np.random.choice(point_range, num_samples, replace=replace)
+        if sample_range and not replace:
```
Wuziyi616, line 872: Would it be better if we use `if sample_range is not None`? After all it's possible value is `int` or `None`. I understand that `if None` is equal to `if False` but I think `not None` is a better style?

tests/test_data/test_pipelines/test_indoor_sample.py
```diff
@@ -1,14 +1,13 @@
 import numpy as np
 
 from mmdet3d.core.points import DepthPoints
-from mmdet3d.datasets.pipelines import (IndoorPatchPointSample,
-                                        IndoorPointSample,
+from mmdet3d.datasets.pipelines import (IndoorPatchPointSample, PointSample,
                                         PointSegClassMapping)
 
 
 def test_indoor_sample():
```
Wuziyi616, line 8: Maybe we should add a case to test `PointSample` with `sample_range is not None`? After all, this is a newly introduced function.

wHao-Wu: Thanks for your advice. However, there needs a data conversion from LiDAR to CAMERA or DEPTH before PointSample, which is not supported by the current repo. After a future merge of function 'data conversion', I will add an additional unittest.

Wuziyi616: But I guess in unit tests, we can manually construct some Camera points and use them as input? I don't think it's a good idea to 'tie' two transforms together and only test them together? Separately testing makes fewer bugs I believe.

mmdet3d/datasets/pipelines/transforms_3d.py
wHao-Wu: Sorry for the late reply. I accidentally missed this review before.

If we sample points from point clouds containing few points, then the sample range will not take effect. And all points, including far points and near points, need to be sampled with the option `replace`.

wHao-Wu: Sorry for the late reply. I accidentally missed this review before.

If we sample points from point clouds containing few points, then the sample range will not take effect. And all points, including far points and near points, need to be sampled with the option replace.

tests/test_data/test_pipelines/test_indoor_sample.py
wHao-Wu: I just added an unittest in `test_transforms_3d.py` to test the point sampling function.

mmdet3d/datasets/pipelines/transforms_3d.py
Wuziyi616: Oh, I mean changing it to `if sample_range  is not None and not replace:` . My concern is that `sample_range` is not a `bool` variable, so directly using `if` on it is not standard?

tests/test_data/test_pipelines/test_indoor_sample.py
Wuziyi616: Yes, great!

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,61 @@ def __repr__(self):
...
-
                 - points (np.ndarray | :obj:`BasePoints`): 3D Points.
                 - choices (np.ndarray, optional): The generated random samples.
         """
         if replace is None:
             replace = (points.shape[0] < num_samples)
-        choices = np.random.choice(
-            points.shape[0], num_samples, replace=replace)
+        point_range = range(len(points))
+        if sample_range:
```
Wuziyi616, line 863: Also change this to `if sample_range is not None` would be better?

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -867,14 +883,19 @@ def __call__(self, results):
...
             dict: Results after sampling, 'points', 'pts_instance_mask' \
                 and 'pts_semantic_mask' keys are updated in the result dict.
         """
+        from mmdet3d.core.points import CameraPoints
         points = results['points']
-        points, choices = self.points_random_sampling(
-            points, self.num_points, return_choices=True)
+        # Points in Camera coord can provide the depth information.
+        # TODO: Need to suport distance-based sampling for other coord system.
+        if self.sample_range:
```
Wuziyi616, line 894: Also change to `if self.sample_range is not None` ?

wHao-Wu: Comments resolved. Code refined. Thx.

Wuziyi616: This line is not resolved yet?

Wuziyi616 (approved):
Almost ready to merge with one last comment

mmdet3d/datasets/pipelines/transforms_3d.py
wHao-Wu: That's just a reminder for developers that we can only support sampling with distance for camera coord system. If necessary, I will switch it to become a `NOTE` message instead of a `TODO` message.

wHao-Wu: Comments resolved. Thx.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,60 @@ def __repr__(self):
...
                 - choices (np.ndarray, optional): The generated random samples.
         """
         if replace is None:
             replace = (points.shape[0] < num_samples)
-        choices = np.random.choice(
-            points.shape[0], num_samples, replace=replace)
+        point_range = range(len(points))
+        if sample_range is not None and not replace:
+            # Only sampling the near points when len(points) >= num_samples
+            depth = points.coord[:, 2]
```
xiliu8006, line 865: why is the depth = points.coords[:, 2] and not depth=np.linalg.norm(points_numpy[:, 0:3], axis=1) with the points_numpy = points.tensor.clone().numpy()

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,60 @@ def __repr__(self):
...
         Args:
             points (np.ndarray | :obj:`BasePoints`): 3D Points.
             num_samples (int): Number of samples to be sampled.
-            replace (bool): Whether the sample is with or without replacement.
-            Defaults to None.
-            return_choices (bool): Whether return choice. Defaults to False.
-
+            sample_range (int, optional): Indicating the range where the points
+                will be sampled.
+                Defaults to None.
```
xiliu8006, line 850: why the type of sample_range is int?  if we use it as a threshold to get the far_inds and near_inds, may use the float is better?

wHao-Wu: Sorry for that. It's reasonable to switch it to float.

mmdet3d/datasets/pipelines/transforms_3d.py
wHao-Wu: In the original source code of point rcnn, it is true that using the third dimensions to be as the depths of points. I also think it is unreasonable. And the latest code will change the behavior when calculating depth.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,60 @@ def __repr__(self):
...
+        point_range = range(len(points))
+        if sample_range is not None and not replace:
+            # Only sampling the near points when len(points) >= num_samples
+            depth = np.linalg.norm(points.tensor, axis=1)
+            far_inds = np.where(depth > sample_range)[0]
+            near_inds = np.where(depth <= sample_range)[0]
+            point_range = near_inds
+            num_samples -= len(far_inds)
+        choices = np.random.choice(point_range, num_samples, replace=replace)
+        if sample_range is not None and not replace:
```
xiliu8006, line 871: this is necessary？ it is already done in L863

xiliu8006 (approved):

mmdet3d/datasets/pipelines/transforms_3d.py
wHao-Wu: The goals of two `if` statements are different. In the first one, we can determine the sampling range. In the second one, we can combine the sampled near points and far points.

mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -818,45 +818,60 @@ def __repr__(self):
...
-                               num_samples,
-                               replace=None,
-                               return_choices=False):
+        self.sample_range = sample_range
+
+    def _points_random_sampling(self,
+                                points,
+                                num_samples,
+                                sample_range=None,
+                                replace=None,
```
ZwwWayne, line 839: 1. by default, it should be False? 
2. It seems that it cannot be set from the config/module and it is not used in __call__, so should we keep it?

wHao-Wu: I think `replace` is useful especially when in case that users want to use replace explicitly. So I create an interface for `replace`.

wHao-Wu: Besides, thanks for your remind, I found that most of `replace` in `mmdet3d/datasets/pipelines/transforms_3d.py` are not used in **call**. So there is a choice: keeping shutdown or open for all `replace`.

ZwwWayne:
Can be merged after resolving conflicts

PR #695, opened 2021-07-03
[Enhance] Support RGB images on ScanNet for multi-view detector

filaPro:
This PR aims to support RGB images on ScanNet for multi-view detector. More discussion in #620.

It is a bit WIP now as I also plan to add images to `ScanNetDataset`. For now I just adapted `SensReader` from original ScanNet [repo](https://github.com/ScanNet/ScanNet/tree/master/SensReader/python). Can you please check if we are ok with it? cc @Tai-Wang


filaPro:
Sorry, attached wrong files. Will close and open another PR.

PR #696, opened 2021-07-03
[Enhance] Support RGB images on ScanNet for multi-view detector

filaPro:
This PR aims to support RGB images on ScanNet for multi-view detector. More discussion in #620.

It is a bit WIP now as I also plan to add images to `ScanNetDataset`. For now I just adapted `SensReader` from original ScanNet [repo](https://github.com/ScanNet/ScanNet/tree/master/SensReader/python). Can you please check if we are ok with it? cc @Tai-Wang

### DIFF
--- a/data/scannet/README.md
+++ b/data/scannet/README.md
@@ -6,7 +6,7 @@ We follow the procedure in [votenet](https://github.com/facebookresearch/votenet
 
 2. In this directory, extract point clouds and annotations by running `python batch_load_scannet_data.py`. Add the `--max_num_point 50000` flag if you only use the ScanNet data for the detection task. It will downsample the scenes to less points.
 
-3. In this directory, extract RGB image with poses by running `python extract_posed_images.py`. This step is optional. Skip it if you don't plan to use multi-view RGB images. Add `--max-images-per-scene -1` to disable limiting number of images per scene. ScanNet scenes contain up to 5000+ frames per each. And after extraction .jpg images requires around 2 Tb of dist space. The recommended 300 images per scene requires less then 100 Gb. For example multi-view 3d detector ImVoxelNet samples 50 and 100 images per training and test scene.
+3. In this directory, extract RGB image with poses by running `python extract_posed_images.py`. This step is optional. Skip it if you don't plan to use multi-view RGB images. Add `--max-images-per-scene -1` to disable limiting number of images per scene. ScanNet scenes contain up to 5000+ frames per each. After extraction, all the .jpg images require 2 Tb disk space. The recommended 300 images per scene require less then 100 Gb. For example multi-view 3d detector ImVoxelNet samples 50 and 100 images per training and test scene.
 
 4. Enter the project root directory, generate training data by running
 
@@ -18,6 +18,7 @@ The overall process could be achieved through the following script
 
 ```bash
 python batch_load_scannet_data.py
+python extract_posed_images.py 
 cd ../..
 python tools/create_data.py scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
 ```
@@ -45,6 +46,11 @@ scannet
 │   ├── train_resampled_scene_idxs.npy
 │   ├── val_label_weight.npy
 │   ├── val_resampled_scene_idxs.npy
+├── posed_images
+│   ├── scenexxxx_xx
+│   │   ├── xxxxxx.txt
+│   │   ├── xxxxxx.jpg
+│   │   ├── intrinsic.txt
 ├── scannet_infos_train.pkl
 ├── scannet_infos_val.pkl
 ├── scannet_infos_test.pkl
--- a/data/scannet/README.md
+++ b/data/scannet/README.md
@@ -6,7 +6,9 @@ We follow the procedure in [votenet](https://github.com/facebookresearch/votenet
 
 2. In this directory, extract point clouds and annotations by running `python batch_load_scannet_data.py`. Add the `--max_num_point 50000` flag if you only use the ScanNet data for the detection task. It will downsample the scenes to less points.
 
-3. Enter the project root directory, generate training data by running
+3. In this directory, extract RGB image with poses by running `python extract_posed_images.py`. This step is optional. Skip it if you don't plan to use multi-view RGB images. Add `--max-images-per-scene -1` to disable limiting number of images per scene. ScanNet scenes contain up to 5000+ frames per each. And after extraction .jpg images requires around 2 Tb of dist space. The recommended 300 images per scene requires less then 100 Gb. For example multi-view 3d detector ImVoxelNet samples 50 and 100 images per training and test scene.
+
+4. Enter the project root directory, generate training data by running
 
 ```bash
 python tools/create_data.py scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
--- a/data/scannet/extract_posed_images.py
+++ b/data/scannet/extract_posed_images.py
@@ -19,7 +19,7 @@ COMPRESSION_TYPE_DEPTH = {
 
 
 class RGBDFrame:
-
+    """Class for single ScanNet RGB-D image processing."""
     def load(self, file_handle):
         self.camera_to_world = np.asarray(
             struct.unpack('f' * 16, file_handle.read(16 * 4)),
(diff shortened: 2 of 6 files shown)


data/scannet/README.md
```diff
@@ -6,7 +6,9 @@ We follow the procedure in [votenet](https://github.com/facebookresearch/votenet
 
 2. In this directory, extract point clouds and annotations by running `python batch_load_scannet_data.py`. Add the `--max_num_point 50000` flag if you only use the ScanNet data for the detection task. It will downsample the scenes to less points.
 
-3. Enter the project root directory, generate training data by running
+3. In this directory, extract RGB image with poses by running `python extract_posed_images.py`. This step is optional. Skip it if you don't plan to use multi-view RGB images. Add `--max-images-per-scene -1` to disable limiting number of images per scene. ScanNet scenes contain up to 5000+ frames per each. And after extraction .jpg images requires around 2 Tb of dist space. The recommended 300 images per scene requires less then 100 Gb. For example multi-view 3d detector ImVoxelNet samples 50 and 100 images per training and test scene.
```
Tai-Wang, line 9: And after extraction .jpg images requires 2 Tb of dist space. -> After extraction, all the .jpg images require 2 Tb disk space.
The recommended 300 images ... requires -> require

Tai-Wang (commented):
We would better add some basic comments or docstrings to `extract_posed_images.py` and add the corresponding contents/explanations in the [ScanNet dataset doc](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/datasets/scannet_det.md).

Tai-Wang:
> This PR aims to support RGB images on ScanNet for multi-view detector. More discussion in #620.
> 
> It is a bit WIP now as I also plan to add images to `ScanNetDataset`. For now I just adapted `SensReader` from original ScanNet [repo](https://github.com/ScanNet/ScanNet/tree/master/SensReader/python). Can you please check if we are ok with it? cc @Tai-Wang

It is fine to borrow some official code from ScanNet with the explicit acknowledgement/citation to their credit. Just need some basic comments and explanations in the doc to make this new feature clearer.

filaPro:
Hi @Tai-Wang ,

Have a question about `LoadMultiViewImageFromFiles`. Looks like it is now not used in the project and also in my opinion it should be replaced with another class. The point is how to apply per image transforms like `Resize` or `Normalize` to the output of this class? It probably will cause `MultiViewResize` and `MultiViewNormalize` which looks not good. I think we should use the idea of `MultiScaleFlipAug3D(..., transforms=[...]) ` in the way of `MultiViewPipeline(..., transforms=[..., Resize, Normalize])`. Do you agree or see a better solution?

Also a bit curious about your `LoadImageFromFileMono3D`. For some reason in all other datasets we return calibration matrices directly in dataset, and for `NuScenesMonoDataset` we have a standalone pipeline. Can we move this matrix to `NuScenesMonoDataset.pre_pipeline`?

Tai-Wang:
Hi @filaPro ,

`LoadMultiViewImageFromFiles` is a pipeline designed for MVXNet on nuScenes. Because the release of moca is blocked, it has not been used in current supported models. You can basically keep it for further potential usage and add your pipelines. I agree that following the approach of `MultiScaleFlipAug3D` to achieve transformations to multi-view images is a good choice.

As for `LoadImageFromFileMono3D`, it is just because when first designed this pipeline I add some other functions, and at last only cam_intrinsic information is preserved. I agree that moving it into functions in the datasets like `get_ann_infos` or `parse_ann_infos` would be a good idea to make it consistent with other datasets and unify the pipelines. I will consider to refactor it before next release.

filaPro:
Hi @Tai-Wang ,
I think it is enough for this PR. `MultiViewPypeline` can be committed in a separate one. I've also visualized a couple of boxes on RGB images, looks fine.

docs/datasets/scannet_det.md
```diff
@@ -127,6 +127,16 @@ def export(mesh_file,
...
 
+### Export ScanNet RGB data
+
+By exporting ScanNet RGB data, for each scene we load a set of RGB images with corresponding 4x4 pose matrices, and a single 4x4 camera intrinsic matrix. Note, that this step is optional and can be skipped if multi-view detection is not planned to use.
+
+```shell
+python extract_posed_images.py
+```
+
+Each of 1201 train, 312 validation and ??? test scenes contains a single `.sens` file. For instance, for scene `0001_01` we have `data/scannet/scans/scene0001_01/0001_01.sens`. For this scene all images and poses are extracted to `data/scannet/posed_images/scene0001_01`. Specifically, there will be 300 image files xxxxx.jpg, 300 camera pose files xxxxx.txt and a single `intrinsic.txt` file. Typically, single scene contains several thousand images. By default, we extract only 300 of them with resulting weight of <100 Gb. To extract more images use `--max-images-per-scene` parameter.
```
Tai-Wang, line 138: ??? test scenes?

Tai-Wang: Typically, single scene -> Typically, a single scene
To extract more images use -> To extract more images, use

Tai-Wang (commented):
Almost ready to be merged

Tai-Wang (approved):

Wuziyi616 (approved):
@filaPro Good job! From the ImVoxelNet paper, it seems to perform training on training set and report results on val set. So all the gt boxes it uses is axis-aligned, and test set is not involved right? If that's true, this PR is done.

I just want to make sure this point because we perform point cloud alignment in the data loading pipeline. Gt boxes in both aligned and unaligned coordinates are provided.

filaPro:
Sure, in ImVoxelNet paper only train and val scenes are used. 

Btw, not quite understand how train/val/test splits are connected with point cloud alignment? We can not process this alignment on test step if necessary?

Wuziyi616:
> 
> 
> Sure, in ImVoxelNet paper only train and val scenes are used.
> 
> Btw, not quite understand how train/val/test splits are connected with point cloud alignment? We can not process this alignment on test step if necessary?

Good question. This is mainly because the **axis-aligned matrix** required to perform alignment is consider as part of the annotation by ScanNet team. As a result, test scans don't have this matrix and can't be aligned. This [doc](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/compatibility.md#scannet-data-pre-processing) provide more details about our design of ScanNet pre-processing scripts (which I believe is a bit complicated and confusing at the first glance).

BTW it reminds me that, it would be great if you can add this PR's change in the compatibility doc, so that users will know they may need to re-generate ScanNet data for your ImVoxelNet models (and maybe more multi-view method in the future :)

filaPro:
Can I add a `MMDetection3D 0.16.0` section in this `compatibility.md`?

Wuziyi616:
> 
> 
> Can I add a `MMDetection3D 0.16.0` section in this `compatibility.md`?

Yes, sure! Just follow the pattern :)

docs/compatibility.md
```diff
@@ -2,6 +2,18 @@
 
 This document provides detailed descriptions of the BC-breaking changes in MMDetection3D.
 
+## MMDetection3D 0.16.0
+
+### ScanNet dataset for ImVoxelNet
+
+We adopt a new pre-processing procedure for the ScanNet dataset in order to support ImVoxelNet, which is a multi-view method requiring image data. In previous versions of MMDetection3D, ScanNet dataset was only used for point cloud based 3D detection methods. We plan adding ImVoxelNet to our model zoo, thus updating ScanNet correspondingly by adding image-related pre-processing steps. Specifically, we made these changes:
```
Wuziyi616, line 9: `ScanNet dataset was only used for point cloud based 3D detection methods` --> `ScanNet dataset was only used for point cloud based 3D detection **and segmentation** methods`

Wuziyi616:
Ready to merge after fixing that typo.

data/scannet/extract_posed_images.py
```diff
@@ -0,0 +1,169 @@
...
+    print(f'processing {idx}')
+    data = SensorData(os.path.join(path, idx, f'{idx}.sens'), limit)
+    output_path = os.path.join('posed_images', idx)
+    data.export_color_images(output_path)
+    data.export_intrinsics(output_path)
+    data.export_poses(output_path)
+
+
+def process_directory(path, limit):
+    with Pool(8) as pool:
```
ZwwWayne, line 155: May allow setting the number of workers by an argument like `--nproc`.
To make users able to feel the process, may also consider use mmcv.track_parallel_process

PR #705, opened 2021-07-06
[Enhance] Modify the Fcos3d init_weights methods

xiliu8006:
## Motivation
Replace the init_weights() with init_cfg.

## Modification
Modified the Fcos3d init_cfg


Tai-Wang (approved):

mmdet3d/models/dense_heads/anchor_free_mono3d_head.py
```diff
@@ -266,29 +273,8 @@ def _init_predictor(self):
...
-        if self.use_direction_classifier:
-            for m in self.conv_dir_cls_prev:
-                if isinstance(m.conv, nn.Conv2d):
-                    normal_init(m.conv, std=0.01)
-        if self.pred_attrs:
-            for m in self.conv_attr_prev:
-                if isinstance(m.conv, nn.Conv2d):
-                    normal_init(m.conv, std=0.01)
+        super().init_weights()
         bias_cls = bias_init_with_prob(0.01)
```
ZwwWayne, line 277: We can also modify the bias through init_cfg

ZwwWayne: by override

xiliu8006: We are not sure the self.use_direction_classifier and self.pred_attrs are always True, So we can not use the override.

PR #709, opened 2021-07-06
[Fix] Convert rt_mat to torch.Tensor in Coord transform for compatibility

### LINKED ISSUE #707: Possible bug with 'rt_mat (np.ndarray)' in coord_3d_mode
https://github.com/open-mmlab/mmdetection3d/blob/a1e04538295b02d42fc59bb7ed58c59c98220174/mmdet3d/core/bbox/structures/coord_3d_mode.py#L246

Should it be `if rt_mat.shape[1] == 4:` to take into account both (np.ndarray | torch.Tensor) types of `rt_mat` as input 
or
convert `rt_mat` from np.ndarray to torch.Tensor before using `if rt_mat.size(1) == 4` ?

gopi231091:

## Motivation
Convert `rt_mat` in 'coord_3d_mode.py' from np.ndarray to torch.Tensor. Closes #707 

## Modification
The `rt_mat` in 'coord_3d_mode.py' takes np.ndarray and torch.tensor types, but the conversion from np.ndarray to torch.Tensor is missing as further code assumes that `rt_mat` is a torch.Tensor type. So, conversion of `rt_mat` from np.ndarray to torch.tensor type is added! 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet3d/core/bbox/structures/coord_3d_mode.py
+++ b/mmdet3d/core/bbox/structures/coord_3d_mode.py
@@ -243,6 +243,9 @@ class Coord3DMode(IntEnum):
                 f'Conversion from Coord3DMode {src} to {dst} '
                 'is not supported yet')
 
+        if isinstance(rt_mat, np.ndarray):
+            rt_mat = arr.new_tensor(rt_mat)
+
         if rt_mat.size(1) == 4:
             extended_xyz = torch.cat(
                 [arr[:, :3], arr.new_ones(arr.size(0), 1)], dim=-1)
--- a/mmdet3d/core/bbox/structures/coord_3d_mode.py
+++ b/mmdet3d/core/bbox/structures/coord_3d_mode.py
@@ -243,9 +243,8 @@ class Coord3DMode(IntEnum):
                 f'Conversion from Coord3DMode {src} to {dst} '
                 'is not supported yet')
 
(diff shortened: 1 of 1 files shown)


mmdet3d/core/bbox/structures/coord_3d_mode.py
```diff
@@ -243,6 +243,9 @@ def convert_point(point, src, dst, rt_mat=None):
                 f'Conversion from Coord3DMode {src} to {dst} '
                 'is not supported yet')
 
+        if isinstance(rt_mat, np.ndarray):
```
Wuziyi616, line 246: Can we modify the if condition to be consistent with [here](https://github.com/open-mmlab/mmdetection3d/blob/a1e04538295b02d42fc59bb7ed58c59c98220174/mmdet3d/core/bbox/structures/coord_3d_mode.py#L146)? After all, they are in the same file. It would be better to have a consistent style.

Wuziyi616 (commented):
Only need to fix that small comment and wait for CI to pass.

mmdet3d/core/bbox/structures/coord_3d_mode.py
gopi231091: Done!

Wuziyi616 (approved):

PR #744, opened 2021-07-13
[Fix] Fix the dimension and yaw hack of FCOS3D on nuScenes

Tai-Wang:

## Motivation
Due to inconsistent definition of nuscbox and our cambox, the transformation of dimension and yaw is required in the pre-processing and post-processing.

## Modification
1. Dimension and yaw transformed to our cambox in the pre-processing.
2. Corresponding transformation in the post-processing functions.
3. Because the yaw angle defined in our cambox is the global yaw instead of local yaw, we move the transformation in the kitti/nuscenes mono dataset into get_target_3d of fcos3d head.
4. Fix the minor errors caused by the change of cam_intrinsic when applying RandomFlip3D.
5. Remove the hack in the mono3d visualization code.

## BC-breaking (Optional)
Json files of nuScenes and KITTI datasets need to be regenerated.

## Use cases (Optional)

## Checklist


mmdet3d/datasets/pipelines/transforms_3d.py
```diff
@@ -119,9 +119,11 @@ def random_flip_data_3d(self, input_dict, direction='horizontal'):
         if 'centers2d' in input_dict:
             assert self.sync_2d is True and direction == 'horizontal', \
                 'Only support sync_2d=True and horizontal flip with images'
-            w = input_dict['img_shape'][1]
+            w = input_dict['ori_shape'][1]
             input_dict['centers2d'][..., 0] = \
                 w - input_dict['centers2d'][..., 0]
+            input_dict['cam_intrinsic'][0][2] = \
+                w - input_dict['cam_intrinsic'][0][2]
```
Wuziyi616, line 126: Can we add a comment to explain this? This is a bit hard to understand at first glance for beginners.

Wuziyi616 (commented):
Good job! Can we verify the correctness by doing some visualizations? It would be great if you can post results in demo, browse_dataset and visualization while doing model inference below.

Wuziyi616:
I think we also need to update FCOS3D page with new checkpoints, and modify the compatibility docs.

Tai-Wang:
> I think we also need to update FCOS3D page with new checkpoints, and modify the compatibility docs.

Yes, that's why I add a label WIP :)

Tai-Wang:
An example to illustrate the correction of camera intrinsic matrix in the RandomFlip3D.
The original image and the corresponding 3D bounding box:
image
Our image flip is typically based on the horizontal center (e.g. a 1600x900 image is flipped around u=800 pixels). However, the horizontal position of camera center in the image is typically not strictly equal to 800 (for example, sometimes 780 or 820), so it will yield errors when flipping images without manipulating the camera intrinsic matrix accordingly.
Without modifying the camera intrinsics, the visualization will be:
image
With our correction, it will be:
image
So we can see the necessity of modifying the camera intrinsics and have an intuitive feeling of the yielded error from this example.

mmdet3d/datasets/pipelines/transforms_3d.py
Tai-Wang: Done.

THU17cyz (approved):
Awesome! After this pr is merged I will check on visualization on the coordinate system refactor branch.

THU17cyz:
> Awesome! After this pr is merged I will check on visualization on the coordinate system refactor branch.

Oh sorry, I should not have approved since there is still work going on.

Isaac-xie:
@Tai-Wang Hi, I'm confused about this modification. In transfroms_3d.py I don't see any changes about any parameters, the parameters that were changed are only centers_2d and cam2img.
First, the ux of cam2img is changed, but the change of  ux only causes the image to be panned and not flipped.
Secondly, after the image is flipped, the yaw angle is changed, doesn't this parameter need to be modified?

Isaac-xie:
Hi, I'm confused about this modification. In transfroms_3d.py I only see changes about centers_2d and cam2img.
after the image is flipped, the yaw angle is changed, doesn't this parameter need to be modified?

PR #702, opened 2021-07-06
add groupfree3d-sunrgbd config and evaluate with axis-aligned length/…

hjin2902:
…width

Thanks for your contribution and we appreciate it a lot. The following instructions would make your pull request more healthy and more easily get feedback. If you do not understand some items, don't worry, just make the pull request and seek help from maintainers.

## Motivation
In Group-Free-3D opensource code, axis-aligned length/width rather than real length/width is predicted and evaluated. So here we also provide this optional choice if necessary.

## Modification
1. Add parameter "axis_aligned_lw" in groupfree3d_bbox_coder.py, custom_3d.py and indoor_eval.py
2. Provide config for the training of Group-Free-3D on sunrgbd dataset

## BC-breaking (Optional)
Does the modification introduce changes that break the back-compatibility of the downstream repos?
If so, please describe how it breaks the compatibility and how the downstream projects should modify their code to keep compatibility with this PR.
No

## Use cases (Optional)
If this PR introduces a new feature, it is better to list some use cases here, and update the documentation.

## Checklist

1. Pre-commit or other linting tools are used to fix the potential lint issues.
2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
3. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects.
4. The documentation has been modified accordingly, like docstring or example tutorials.

PR #765, opened 2021-07-19
[Fix] Add `max_keep_ckpts` for groupfree3d

THU17cyz:

## Motivation
The size of a most simple groupfree3d ckpt is around 170MB, and there are by default 400 ckpts saved in one run, which will take about 70GB space. There is a high chance of disk space/quota exceed, which will terminate training.

## Modification
I added  `max_keep_ckpts`  in groupfree3d configs.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


THU17cyz:
@hjin2902 , please take a look. If you approve, you can also update #702 ?

ZwwWayne (approved):

hjin2902:
> @hjin2902 , please take a look. I cannot request review from you. If you approve, you can also update #702 ?

Sorry for the late reply! 
Yes, we can add config like this, but a potential problem is that the best checkpoint may not in the last 10 epoch. So maybe the value of max_keep_ckpts should be large enough.

PR #551, opened 2021-05-17
[Enhance] Adds windows compilation support

divyanshj16:
fixes #169 
Here is the summary of the fixes,
- In windows' compiler variables cannot be passed in the size of initialization of the array, so we have to initialize it something like this. `int *xx= new int[variable]`  - [source](https://stackoverflow.com/questions/56972857/cuda-code-compile-on-linux-but-not-in-windows-visual-studio-2012)
- The long datatype in windows and linux are different so the following source suggested it to convert everything to `long long` .[source](https://forums.developer.nvidia.com/t/strange-link-error-seen-by-multiple-people-while-building-pytorch-cpp-cuda-extensions/145261)
- While defining EPS, its scope had to be changed to `__device__`, otherwise it was throwing errors about undefined EPS.

PR #766, opened 2021-07-19
[Fix] Fix compiling error on windows

### LINKED ISSUE #169: compilation error C2131: expression did not evaluate to a constant
Hi, I'm trying to install and use mmdetection on my Windows 10 machine (I know, that's my first problem...) and I'm at literally the last step of the process. Everything appears to have installed and compiled correctly up to this point, but when running the "pip install -v -e ." inside the mmdetection3d directory I receive the following errors for iou3d:

```
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(108): error C2131: expression did not evaluate to a constant
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(108): note: failure was caused by a read of a variable outside its lifetime
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(108): note: see usage of 'boxes_num'
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(127): error C2131: expression did not evaluate to a constant
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(108): note: failure was caused by a read of a variable outside its lifetime
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(108): note: see usage of 'boxes_num'
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(181): error C2131: expression did not evaluate to a constant
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(162): note: failure was caused by a read of a variable outside its lifetime
D:\Code\3DDetection\mmdetection3d\mmdet3d\ops\iou3d\src\iou3d.cpp(162): note: see usage of 'boxes_num'

```
In researching this particular error, it appears to be a problem with the compiler in Windows. The problem is line 108 mentioned in the above errors:

`const int col_blocks = DIVUP(boxes_num, THREADS_PER_BLOCK_NMS);`

and then its use on line 127 to initialize an array:

`unsigned long long remv_cpu[col_blocks];`

Apparently some compilers are okay with this and some, specifically in Windows apparently, are not. I do not know how to work around this problem. I have tried various things mentioned in places like StackOverflow, like setting col_blocks to be static, but the same error appears. I have also checked the issue queue for OpenPCDet, where the original code for iou3d came from, but nobody seems to have this problem. Does anyone know of a way around this?

wangruohui:
Fix #169, related to #551 , but the author does not respond.

### DIFF
--- a/mmdet3d/ops/iou3d/src/iou3d.cpp
+++ b/mmdet3d/ops/iou3d/src/iou3d.cpp
@@ -103,7 +103,7 @@ int nms_gpu(at::Tensor boxes, at::Tensor keep,
 
   int boxes_num = boxes.size(0);
   const float *boxes_data = boxes.data_ptr<float>();
-  long *keep_data = keep.data_ptr<long>();
+  long long *keep_data = keep.data_ptr<long long>();
 
   const int col_blocks = DIVUP(boxes_num, THREADS_PER_BLOCK_NMS);
 
@@ -124,8 +124,7 @@ int nms_gpu(at::Tensor boxes, at::Tensor keep,
 
   cudaFree(mask_data);
 
-  unsigned long long remv_cpu[col_blocks];
-  memset(remv_cpu, 0, col_blocks * sizeof(unsigned long long));
+  unsigned long long* remv_cpu = new unsigned long long[col_blocks]();
 
   int num_to_keep = 0;
 
@@ -141,6 +140,7 @@ int nms_gpu(at::Tensor boxes, at::Tensor keep,
       }
     }
   }
+  delete[] remv_cpu;
   if (cudaSuccess != cudaGetLastError()) printf("Error!\n");
 
   return num_to_keep;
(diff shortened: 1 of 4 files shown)


Wuziyi616:
@wangruohui Seems that we can successfully build these ops, but will cause errors when importing them. Can you successfully train models using these modified ops in your Linux server?

Wuziyi616:
Hi @wangruohui, can you try re-training one model which uses these ops to benchmark the result? For example, [`PartA2Net`](https://github.com/open-mmlab/mmdetection3d/blob/master/configs/parta2/hv_PartA2_secfpn_2x8_cyclic_80e_kitti-3d-3class.py) that uses `iou3d`. If you have trouble downloading the data, I can train it for you.

ZwwWayne:
> Hi @wangruohui, can you try re-training one model which uses these ops to benchmark the result? For example, [`PartA2Net`](https://github.com/open-mmlab/mmdetection3d/blob/master/configs/parta2/hv_PartA2_secfpn_2x8_cyclic_80e_kitti-3d-3class.py) that uses `iou3d`. If you have trouble downloading the data, I can train it for you.

Please train one with this PR. Actually, if the related model can test well, we can merge that. The modification should be fine.

Wuziyi616 (approved):
Have tested PointPillars on KITTI, PartA2 on KITTI, VoteNet on ScanNet and PAConv on S3DIS. They cover all the modified CUDA files in this PR. The test accuracy is good.

wangruohui:
Thank you very much!

PR #779, opened 2021-07-22
[Feature] Support PointRCNN in MMDet3D

xiliu8006:
## Motivation
Support PointRCNN in MMDet3D.


mmdet3d/models/detectors/base.py
```diff
@@ -104,4 +104,4 @@ def show_results(self, data, result, out_dir):
                 ValueError(
                     f'Unsupported box_mode_3d {box_mode_3d} for convertion!')
             pred_bboxes = pred_bboxes.tensor.cpu().numpy()
-            show_result(points, None, pred_bboxes, out_dir, file_name)
+            show_result(points, None, pred_bboxes, out_dir, file_name, False)
```
wHao-Wu, line 107: Is it necessary to change the default behavior of `show_results`.

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,66 @@
+from mmcv.runner import BaseModule
+from torch import nn as nn
+
+from mmdet3d.ops import PointFPModule
+from mmdet.models import NECKS
+
+
+@NECKS.register_module()
+class PointNetFPNeck(BaseModule):
```
wHao-Wu, line 9: I'm not sure whether we need to follow the structure of [PointNetSSG](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/models/backbones/pointnet2_sa_ssg.py), which incorporates the FP module into itself.

xiliu8006: @ZwwWayne and I think the FP modules put in necks is better so I create this file.

mmdet3d/models/detectors/base.py
xiliu8006: It is unnecessary, I will modified in a new PR

robin-karlsson0:
@xiliu8006 Thank you for implementing PointRCNN!

Is it alright to ask a question about the implementation here?

The paper states (p.6, Sec. 4.1 Implementation Details, The training scheme)

> The two stage sub-networks of PointRCNN are **trained separately**.

From reading the implementation config and code, it looks like both the stage 1 (Bottom-up 3D Proposal Generation) and stage 2 (Canonical 3D Box Refinement) sub-networks are **trained simultaneously** with the same schedule and training pipeline.

Could you confirm if I'm missing something, or if separate sequential training is found to be unnecessary?

- - -

PS can confirm that I managed to reproduce the performance on KITTI reported in the paper on my system by just running the PR code.

```
Single RTX 3090 experiment
Batch size: 8x1 = 8 samples
LR: 0.001

Comparison with PointRCNN paper results (Moderate 'car' on validation set)

|    [AP] |  Easy | Moderate |  Hard |
--------------------------------------
| MMDet3D | 89.31 | 78.86    | 78.55 |
| Paper   | 88.88 | 78.63    | 77.38 |
--------------------------------------
```

xiliu8006:
> @xiliu8006 Thank you for implementing PointRCNN!
> 
> Is it alright to ask a question about the implementation here?
> 
> The paper states (p.6, Sec. 4.1 Implementation Details, The training scheme)
> 
> > The two stage sub-networks of PointRCNN are **trained separately**.
> 
> From reading the implementation config and code, it looks like both the stage 1 (Bottom-up 3D Proposal Generation) and stage 2 (Canonical 3D Box Refinement) sub-networks are **trained simultaneously** with the same schedule and training pipeline.
> 
> Could you confirm if I'm missing something, or if separate sequential training is found to be unnecessary?
> 
> PS can confirm that I managed to reproduce the performance on KITTI reported in the paper on my system by just running the PR code.
> 
> ```
> Single RTX 3090 experiment
> Batch size: 8x1 = 8 samples
> LR: 0.001
> 
> Comparison with PointRCNN paper results (Moderate 'car' on validation set)
> 
> |    [AP] |  Easy | Moderate |  Hard |
> --------------------------------------
> | MMDet3D | 89.31 | 78.86    | 78.55 |
> | Paper   | 88.88 | 78.63    | 77.38 |
> --------------------------------------
> ```

it is unnecessary to separate the training steps! we reimplement PointRCNN following this [codebase](https://github.com/open-mmlab/OpenPCDet). This codebase is made by the author of PointRCNN.

PR #791, opened 2021-07-26
[Enhance] Update PointFusion

filaPro:
We use `point_sample` in `MVXNet` and in `ImVoxelNet` both on KITTI. We are also working on reimplementing multi-view `ImVoxelNet` in ScanNet. This PR update `point_sample` from only `lidar2img` to `depth2img` or `cam2img`.

@Tai-Wang can you have a look? I tested  `MVXNet` and `ImVoxelNet` checkpoint. Metrics are exactly the same after this PR.


mmdet3d/models/fusion_layers/point_fusion.py
```diff
@@ -293,6 +286,7 @@ def sample_single(self, img_feats, pts, img_meta):
             img_feats,
             pts,
             pts.new_tensor(img_meta['lidar2img']),
+            'LIDAR',
```
ZwwWayne, line 289: A hardcode here might be inappropriate, can we support a new option like coord_mode in this module? Thus users can change it through config

filaPro:
Hi @ZwwWayne ,
I've fixed your comment. Can you please advice me about these 3 [points](https://github.com/open-mmlab/mmdetection3d/blob/57f1c7d472cdd884799ab4b96b3fed473946f978/mmdet3d/models/fusion_layers/point_fusion.py#L92). Can make 3 separate PRs for them.

Tai-Wang:
> Hi @ZwwWayne ,
> 
> I've fixed your comment. Can you please advice me about these 3 [points](https://github.com/open-mmlab/mmdetection3d/blob/57f1c7d472cdd884799ab4b96b3fed473946f978/mmdet3d/models/fusion_layers/point_fusion.py#L92). Can make 3 separate PRs for them.
> 
> BTW, for now the codebase contains `box_type_3d='Camera'`, `box_mode='camera'` and `points_type == 'CAMERA'`, which is very dirty, in my opinion.

I think we can move the function to `mmdet3d/core/bbox/structures/utils.py` or create a new utils.py under `mmdet3d/core/points/` to make it convenient to use in different models? For other suggestions, I agree with you. Feel free to refactor them with PRs.

ZwwWayne:
Hi @filaPro ,
Thanks for the efforts.  Actually, I agree with the three points. Therefore, as the second point of cam_intrinsic has been refactored, this PR might be updated. We can do point 1 first. For the third point, it is indeed beneficial if we can use a similar coord_type for both boxes and points. Since this might take some time and need refactoring, I suggest doing that with v1.0.0.dev0 branch. Which refactored the whole coordinate system.

ZwwWayne (approved):

PR #809, opened 2021-07-30
[Enhance] Benchmark PAConv on S3DIS dataset

Wuziyi616:
Release PAConv model on S3DIS semantic segmentation task.

Just FYI, the paper's result is `66.33%` while ours is `66.65%`.


Wuziyi616 (commented):
@ZwwWayne checkpoint and log need to be uploaded

ZwwWayne (approved):

PR #815, opened 2021-08-02
add roipooling cuda ops

xiliu8006:

## Motivation
Add roipooling cuda ops to support pointrcnn.

### DIFF
--- a/mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
+++ b/mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
@@ -1,0 +1,63 @@
+import torch
+from torch import nn as nn
+
+from mmdet3d import ops
+from mmdet3d.core.bbox.structures import rotation_3d_in_axis
+from mmdet.models.builder import ROI_EXTRACTORS
+
+
+@ROI_EXTRACTORS.register_module()
+class Single3DRoIPointExtractor(nn.Module):
+    """Point-wise roi-aware Extractor.
+
+    Extract Point-wise roi features.
+
+    Args:
+        roi_layer (dict): The config of roi layer.
+    """
+
+    def __init__(self, roi_layer=None):
+        super(Single3DRoIPointExtractor, self).__init__()
+        self.roi_layer = self.build_roi_layers(roi_layer)
+
+    def build_roi_layers(self, layer_cfg):
+        """Build roi layers using `layer_cfg`"""
+        cfg = layer_cfg.copy()
+        layer_type = cfg.pop('type')
+        assert hasattr(ops, layer_type)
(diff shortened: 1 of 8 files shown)


mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from torch import nn as nn
+
+from mmdet3d import ops
+from mmdet3d.core.bbox.structures import rotation_3d_in_axis
+from mmdet.models.builder import ROI_EXTRACTORS
+
+
+@ROI_EXTRACTORS.register_module()
+class Single3DRoIPointExtractor(nn.Module):
```
ZwwWayne, line 10: Can we inherits those in MMDet

mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
```diff
@@ -0,0 +1,63 @@
...
+    def build_roi_layers(self, layer_cfg):
+        """Build roi layers using `layer_cfg`"""
+        cfg = layer_cfg.copy()
+        layer_type = cfg.pop('type')
+        assert hasattr(ops, layer_type)
+        layer_cls = getattr(ops, layer_type)
+        roi_layers = layer_cls(**cfg)
+        return roi_layers
+
+    def forward(self, feats, coordinate, batch_inds, rois):
```
ZwwWayne, line 32: We can use torch.no_grad() decorator if the whole forward function does not require gradient.

wHao-Wu:
Maybe we need to modify `setup.py` in order to build the corresponding cuda ops when `pip install -v -e .`.

mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
xiliu8006: BaseROIExtractor need some args we don't use such as `out_channels`, `featmap_strides`. So I think it  is unnecessary

mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py
xiliu8006: it has been used in here[https://github.com/xiliu8006/mmdetection3d/blob/39edd896184417755a6f49be62dd88558d15cc78/mmdet3d/models/roi_heads/roi_extractors/single_roipoint_extractor.py#L47]

xiliu8006:
> Maybe we need to modify `setup.py` in order to build the corresponding cuda ops when `pip install -v -e .`.

I am sorry to forget to add this, I will add it later

PR #858, opened 2021-08-10
[Feature] Support PointRCNN backbone

xiliu8006:
## Motivation
Add pointnet2_fp_neck.py to Support PointRCNN backbone.

## Modification
Modify .


wHao-Wu:
Is it necessary to add `mmdet3d/models/necks/.pointnet2_fp_neck.py.swp `? I think it is just a swap file in Linux.

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+@NECKS.register_module()
+class PointNetFPNeck(BaseModule):
+    """PointNet FP Module used in PointRCNN.
+
+    Refer to the `official code <https://github.com/charlesq34/pointnet2>`_.
+
+    Args:
+        fp_channels (tuple[tuple[int]]): Tuple of mlp channels in FP modules.
```
wHao-Wu, line 15: Need `optional` because the arg has the default value?

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -68,6 +68,8 @@ def __init__(self,
             self.num_point = [num_point]
         elif isinstance(num_point, list) or isinstance(num_point, tuple):
             self.num_point = num_point
+        elif num_point is None:
+            self.num_point = None
```
wHao-Wu, line 72: Maybe we need an additional unittest for the condition when num_point == None.

mmdet3d/models/necks/pointnet2_fp_neck.py
xiliu8006: I this it is not a optional arg, because every different model will have the different setting. The default value is just a example.

mmdet3d/ops/pointnet_modules/point_sa_module.py
xiliu8006: Yes, I will add it later

tests/test_models/test_common_modules/test_pointnet_modules.py
```diff
@@ -217,3 +231,6 @@ def test_pointnet_fp_module():
 
     fp_features = self(xyz1, xyz2, features1, features2)
     assert fp_features.shape == torch.Size([1, 16, 50])
+
+
+test_pointnet_sa_module_msg()
```
Tai-Wang, line 236: Clean

mmdet3d/models/necks/pointnet2_fp_neck.py
Tai-Wang: We just regard an argument optional if a default value is given. If the input value should be carefully set, we should remove the default value, or the error message will be not much straightforward.

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,66 @@
+from mmcv.runner import BaseModule
+from torch import nn as nn
+
+from mmdet3d.ops import PointFPModule
+from mmdet.models import NECKS
+
+
+@NECKS.register_module()
+class PointNetFPNeck(BaseModule):
+    """PointNet FP Module used in PointRCNN.
```
Tai-Wang, line 10: r"""

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,66 @@
...
+        return sa_xyz, sa_features
+
+    def forward(self, feat_dict):
+        """Forward pass.
+
+        Args:
+            feat_dict (dict): Feature dict from backbone.
+
+        Returns:
+            torch.Tensor: neck feature with shape [B, fp_channels[-1][-1], N].
```
Tai-Wang, line 52: Neck feature

Tai-Wang (changes requested):

mmdet3d/models/necks/pointnet2_fp_neck.py
xiliu8006: OK, I got it, I will delete the default value.

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,63 @@
...
+from mmdet3d.ops import PointFPModule
+from mmdet.models import NECKS
+
+
+@NECKS.register_module()
+class PointNetFPNeck(BaseModule):
+    r"""PointNet FP Module used in PointRCNN.
+
+    Refer to the `official code <https://github.com/charlesq34/pointnet2>`_.
+
```
ZwwWayne, line 13: may draw a code block here

ZwwWayne: https://github.com/open-mmlab/mmdetection/blob/master/mmdet/models/roi_heads/mask_heads/fused_semantic_head.py#L15

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,63 @@
...
+        sa_features = feat_dict['sa_features']
+        assert len(sa_xyz) == len(sa_features)
+
+        return sa_xyz, sa_features
+
+    def forward(self, feat_dict):
+        """Forward pass.
+
+        Args:
+            feat_dict (dict): Feature dict from backbone.
```
ZwwWayne, line 46: may indicate what keys are in it

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,63 @@
...
+
+        Returns:
+            torch.Tensor: Neck feature with shape [B, fp_channels[-1][-1], N].
```
ZwwWayne, line 49: wrong type

mmdet3d/models/necks/pointnet2_fp_neck.py
ZwwWayne: Use code-block, otherwise they cannot show properly in the documentation

ZwwWayne: https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/core/bbox/structures/cam_box3d.py#L14

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,76 @@
...
+        self.FP_modules = nn.ModuleList()
+        for cur_fp_mlps in fp_channels:
+            self.FP_modules.append(PointFPModule(mlp_channels=cur_fp_mlps))
+
+    def _extract_input(self, feat_dict):
+        """Extract inputs from features dictionary.
+
+        Args:
+            feat_dict (dict): Feature dict from backbone including points
+                and point features.
```
Tai-Wang, line 42: If the dict does not have many keys, please list them with explanations.

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -110,7 +114,9 @@ def _sample_points(self, points_xyz, features, indices, target_xyz):
         Args:
             points_xyz (Tensor): (B, N, 3) xyz coordinates of the features.
             features (Tensor): (B, C, N) features of each point.
```
Tai-Wang, line 116: torch.Tensor, optional
If `features` is a necessary input variable, please remove the default value.

Tai-Wang (commented):
Just need to fix comments to make this PR ready to merge

mmdet3d/models/necks/pointnet2_fp_neck.py
```diff
@@ -0,0 +1,83 @@
...
+        sa_1 -------------                   |       |
+                          |                  |       |
+        sa_0 -> fp_0 -> fp_module ->fp_1 -> ... -> fp_module -> fp_n
+
+    sa_n including sa_xyz (torch.Tensor) and sa_features (torch.Tensor)
+    fp_n including fp_xyz (torch.Tensor) and fp_features (torch.Tensor)
+
+    Args:
+        fp_channels (tuple[tuple[int]]): Tuple of mlp channels in FP modules.
+    """
```
ZwwWayne, line 29: `init_cfg` is missed in docstring

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -77,8 +79,10 @@ def __init__(self,
         self.fps_mod_list = fps_mod
         self.fps_sample_range_list = fps_sample_range_list
 
-        self.points_sampler = Points_Sampler(self.num_point, self.fps_mod_list,
-                                             self.fps_sample_range_list)
+        self.points_sampler = Points_Sampler(
```
ZwwWayne, line 82: May use a more regular style. The current logic is not reader-friendly. We can simply use
```
if xxx:
    xxx
else:
    xxx

```

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -163,7 +165,6 @@ def forward(self, points):
                 out_sa_xyz.append(sa_xyz[-1])
                 out_sa_features.append(sa_features[-1])
                 out_sa_indices.append(sa_indices[-1])
-
```
Wuziyi616, line 166: Unnecessary change

mmdet3d/models/backbones/pointnet2_sa_msg.py
```diff
@@ -63,7 +63,9 @@ def __init__(self,
         self.out_indices = out_indices
         assert max(out_indices) < self.num_sa
         assert len(num_points) == len(radii) == len(num_samples) == len(
-            sa_channels) == len(aggregation_channels)
+            sa_channels)
+        if aggregation_channels is not None:
```
Wuziyi616, line 67: I just wonder if this change is necessary? You only judge if `aggregation_channels` is `None` here. However in [this line](https://github.com/open-mmlab/mmdetection3d/blob/cab70db5c67178aa90c7bf3aaf1ef844fe39da32/mmdet3d/models/backbones/pointnet2_sa_msg.py#L108), we still index from `aggregation_channels` and it will cause an error if `aggregation_channels is None` right?

In my opinion, I think the `aggregation_channels` should not be `None`. Instead it should be a list of `None`? So we should keep the original assert condition.

mmdet3d/ops/pointnet_modules/point_sa_module.py
Wuziyi616:
```
if self.num_point is not None:
    self.points_sampler = Points_Sampler(xxx)
else:
    self.points_sampler = None
```

mmdet3d/ops/pointnet_modules/point_sa_module.py
```diff
@@ -108,10 +112,12 @@ def _sample_points(self, points_xyz, features, indices, target_xyz):
         Otherwise sample points using `self.points_sampler`.
 
         Args:
-            points_xyz (Tensor): (B, N, 3) xyz coordinates of the features.
-            features (Tensor): (B, C, N) features of each point.
-            indices (Tensor): (B, num_point) Index of the features.
-            target_xyz (Tensor): (B, M, 3) new_xyz coordinates of the outputs.
+            points_xyz (tensor.Tensor): (B, N, 3) xyz coordinates of the
```
Wuziyi616, line 115: Should be `torch.Tensor` instead of `tensor.Tensor`. Also fix the other comments

Wuziyi616 (changes requested):
Maybe we need a unit test for the `PointNetFPNeck`

Tai-Wang (changes requested):

PR #877, opened 2021-08-21
[Fix] ImVotenet config

### LINKED ISSUE #876: Question about ImVoteNet config
These 8 [lines](https://github.com/open-mmlab/mmdetection3d/blob/master/configs/imvotenet/imvotenet_stage2_16x8_sunrgbd-3d-10class.py#L177-L184) in config contains 2 calls of `Flip` and 1 call of `Pad`. First question is why do we need `RandomFlip` call as `RandomFlip3D` does the same thing after all?

Second question, is about the order of `Pad` and `RandomFlip3D`. If I understand `mmdetection` code correctly, we usually first flip the image and then pad it on the bottom and right edges. However here we first pad and then flip. Is it a bug, or it is somehow not important in case of `ImVoteNet`?

Probably, @THU17cyz can help.

filaPro:
Close #876. Fix the order of `Flip` and `Pad` for image. Really, this doesn't improve accuracy.
Before: `mAP_0.25`: 0.64043, `mAP_0.50`: 0.37803.
After: `mAP_0.25`: 0.6397, `mAP_0.50`: 0.3701.
Probably, the difference is only about randomness.

### DIFF
--- a/configs/imvotenet/imvotenet_stage2_16x8_sunrgbd-3d-10class.py
+++ b/configs/imvotenet/imvotenet_stage2_16x8_sunrgbd-3d-10class.py
@@ -174,14 +174,13 @@ train_pipeline = [
     dict(type='LoadAnnotations3D'),
     dict(type='LoadAnnotations', with_bbox=True),
     dict(type='Resize', img_scale=(1333, 600), keep_ratio=True),
-    dict(type='RandomFlip', flip_ratio=0.0),
     dict(type='Normalize', **img_norm_cfg),
-    dict(type='Pad', size_divisor=32),
     dict(
         type='RandomFlip3D',
         sync_2d=False,
         flip_ratio_bev_horizontal=0.5,
     ),
+    dict(type='Pad', size_divisor=32),
     dict(
         type='GlobalRotScaleTrans',
         rot_range=[-0.523599, 0.523599],
@@ -212,9 +211,7 @@ test_pipeline = [
         flip=False,
         transforms=[
             dict(type='Resize', keep_ratio=True),
-            dict(type='RandomFlip', flip_ratio=0.0),
             dict(type='Normalize', **img_norm_cfg),
-            dict(type='Pad', size_divisor=32),
             dict(
                 type='GlobalRotScaleTrans',
                 rot_range=[0, 0],
@@ -225,7 +222,8 @@ test_pipeline = [
                 sync_2d=False,
                 flip_ratio_bev_horizontal=0.5,
             ),
-            dict(type='PointSample', num_points=20000),
+            dict(type='Pad', size_divisor=32),
+            dict(type='IndoorPointSample', num_points=20000),
             dict(
                 type='DefaultFormatBundle3D',
                 class_names=class_names,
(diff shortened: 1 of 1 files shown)


THU17cyz (approved):
Thank you for your contribution! Do we need to update the model as well @ZwwWayne ?

Tai-Wang:
@filaPro @THU17cyz Yes, please also update the model and benchmark. (After all, the `test_pipeline` has been modified so the original checkpoint can also not reproduce the original performance?)

filaPro:
@Tai-Wang ,  as flips are not using in test pipeline (without augmentation) the metrics of current model should not change on the validation set. However, I can share new model and log with worse metrics.

Btw, current metrics are probably not relevant as for #653 .

Tai-Wang:
@filaPro Ah yes, then I guess the original checkpoints are just ok. This PR should be ready.

Tai-Wang (approved):

ZwwWayne:
I think somehow the modification changes the original behavior and we should be careful.
```
    dict(type='RandomFlip', flip_ratio=0.0),
    dict(type='Normalize', **img_norm_cfg),
    dict(type='Pad', size_divisor=32),
    dict(
        type='RandomFlip3D',
        sync_2d=False,
        flip_ratio_bev_horizontal=0.5,
    ),
```
In the original config as above, it means that the image is never flipped during training and the point cloud are flipped with a probability of 0.5, following the official implementation. In such a case, the position of padding does not matter because the image is not flipped.

In the new config in this PR, as `dict(type='RandomFlip', flip_ratio=0.0),` is removed, it means the image and points will all be flipped in an unsynchronized manner, and padding should indeed be put after `RandomFlip3D`. This actually modifies the original behavior though it might not affect the performance. @THU17cyz we may ablate the change and see the influence.

THU17cyz:
> I think somehow the modification changes the original behavior and we should be careful.
> 
> ```
>     dict(type='RandomFlip', flip_ratio=0.0),
>     dict(type='Normalize', **img_norm_cfg),
>     dict(type='Pad', size_divisor=32),
>     dict(
>         type='RandomFlip3D',
>         sync_2d=False,
>         flip_ratio_bev_horizontal=0.5,
>     ),
> ```
> 
> In the original config as above, it means that the image is never flipped during training and the point cloud are flipped with a probability of 0.5, following the official implementation. In such a case, the position of padding does not matter because the image is not flipped.
> 
> In the new config in this PR, as `dict(type='RandomFlip', flip_ratio=0.0),` is removed, it means the image and points will all be flipped in an unsynchronized manner, and padding should indeed be put after `RandomFlip3D`. This actually modifies the original behavior though it might not affect the performance. @THU17cyz we may ablate the change and see the influence.

Yes, I think @ZwwWayne is right. In the original config the image is never flipped because `RandomFlip` with `flip_ratio=0.0` already set the `'flip'` in https://github.com/open-mmlab/mmdetection/blob/e876171cdf6cf09cb45fb7e62051365e35fca167/mmdet/datasets/pipelines/transforms.py#L431, which passes on in the pipeline, causing `RandomFlip3D` to not flip the image. @filaPro

filaPro:
I see, should we close this PR and the issue?

ZwwWayne:
> I see, should we close this PR and the issue?

Yes, I think we can do that. We will update a clarification for that in the documentation.

THU17cyz:
Forgot to say that, the correction of the order of flip and pad is actually useful : )

PR #967, opened 2021-09-26
Dv centerpoint

mickeyouyou:
## Motivation
Dynamic voxelization speed up voxelization and avoid information loss during the quantization process, we want to use it in other method and dataset, so pull this request 

## Modification
detector centerpoint 

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


Tai-Wang:
Hi @mickeyouyou , really sorry for the late review. Could you please fix the problems indicated by linting? In addition, is there any checkpoints and benchmark update for this dynamic voxelization version?

Issue #222, opened 2020-12-09
Visualization of results


SimonDoll:
May someone give me a hint on how to visualize results e.g. pointcloud with predictions and ground truth and maybe a camera image for better visibility?
As stated in the getting_started section I've tried out Meshlab for point cloud / mesh visualization but the generated results from a pretrained network can't be opened by meshlab. Are there any other tools available e.g. open3d?

Tai-Wang:
For 3D visualization, a better solution based on other tools like open3d and mayavi is in our plan.
For 2D visualization, it is also in our plan and for now you can temporarily refer to this [repo](https://github.com/kuixu/kitti_object_vis).
If you would like to use the current visualization function, maybe you can provide more information about why the results can not be opened by meshlab. It should work well from my experience.

PR #979, opened 2021-10-02
ADD Annotation bboxes for output visualization

robin-karlsson0:
This PR adds functionality for test visualization to display ground truth annotation bboxes in addition to prediction bboxes.

For sparse point clouds it can be challenging to visually confirm if a prediction is correct or not. Adding in the ground truth annotation bboxes (blue) in addition to the predicted bboxes (green) greatly improves the interpretability of the model's prediction performance.

This functionality has been requested in prior issues such as https://github.com/open-mmlab/mmdetection3d/issues/222 and can be used without any change in configuration files or CLI arguments.

## Modification
First, the ground truth annotation bboxes for a sample is extracted in the visualization code (mmdet3d/apis/test.py) through the pre-existing data_loader object.

`gt_bboxes = data_loader.dataset.get_ann_info(i)['gt_bboxes_3d']`

Secondly, the `show_results()` function in all detection model takes an additional optional argument for the gt_bboxes object.

`model.module.show_results(data, result, out_dir=out_dir, gt_bboxes=gt_bboxes)`

 If no gt_bboxes object is supplied, the default None value results in the original function execution flow.

Finally, the `show_multi_modality_result()` function is already compatible with the gt_bboxes object and successfully visualizes the annotations.

## Visualization example

Model: CenterPoint
Dataset: KITTI
Predictions: Green
GT annotations: Blue

Screenshot from 2021-10-02 15-02-42
Screenshot from 2021-10-02 15-41-54

### DIFF
--- a/mmdet3d/apis/test.py
+++ b/mmdet3d/apis/test.py
@@ -44,7 +44,9 @@ def single_gpu_test(model,
             models_3d = (Base3DDetector, Base3DSegmentor,
                          SingleStageMono3DDetector)
             if isinstance(model.module, models_3d):
-                model.module.show_results(data, result, out_dir=out_dir)
+                gt_bboxes = data_loader.dataset.get_ann_info(i)['gt_bboxes_3d']
+                model.module.show_results(
+                    data, result, out_dir=out_dir, gt_bboxes=gt_bboxes)
             # Visualize the results of MMDetection model
             # 'show_result' is MMdetection visualization API
             else:
--- a/mmdet3d/models/detectors/base.py
+++ b/mmdet3d/models/detectors/base.py
@@ -60,13 +60,14 @@ class Base3DDetector(BaseDetector):
         else:
             return self.forward_test(**kwargs)
 
-    def show_results(self, data, result, out_dir):
+    def show_results(self, data, result, out_dir, gt_bboxes=None):
         """Results visualization.
 
         Args:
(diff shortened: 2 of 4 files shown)


mmdet3d/apis/test.py
```diff
@@ -44,7 +44,9 @@ def single_gpu_test(model,
             models_3d = (Base3DDetector, Base3DSegmentor,
                          SingleStageMono3DDetector)
             if isinstance(model.module, models_3d):
-                model.module.show_results(data, result, out_dir=out_dir)
+                gt_bboxes = data_loader.dataset.get_ann_info(i)['gt_bboxes_3d']
```
Wuziyi616, line 47: Please assign `gt_bboxes = None` if the dataset doesn't have `gt_bboxes_3d` annotation. Can be something like
```
if data_loader.dataset.test_mode:
    gt_bboxes = None
```
See e.g. this [line](https://github.com/open-mmlab/mmdetection3d/blob/e5a87f35ad65c2a2c31bf6888293bfef1b963bed/mmdet3d/datasets/kitti_dataset.py#L133)

Wuziyi616: Another problem is that, if the model being tested is a 3D segmentor or a multi-modality detector. May need to deal with different situations here (e.g. for segmentor, its GT is `pts_semantic_mask`)

PR #1213, opened 2022-02-08
[Fix] evaluation bug

zeyu-hello:
Thought this couple lines don't change result value, `gt` and `dt` are exchanged to meet function's parameter order.


zeyu-hello:
@ZwwWayne

ZCMax:
I think the original code is OK? Please refer to this previous PR: https://github.com/open-mmlab/mmdetection3d/pull/969.

Issue #296, opened 2021-02-02
frame's timestamp value loses its precision when cast from int64 to np.float32 during the conversion of Waymo dataset


zhanggefan:
Hi, 

It looks like a small bug exists here. Waymo dataset uses type int64 for timestamp, but the following code casts its value from int64 to float64 at line 239 (behavior of numpy v1.18.5, though intensity is float32), and at line 247 casts its value further from float64 to float32.

For a typical waymo frame timestamp value, e.g. 1550083467346370, will lose precision and be cast into 1550083493462016. A big value of similar range has a low precision of 134,217,728 (physically 134s) when represented by float32.

This should have little impact though. But I found that the output waymo_infos_*.pkl files are affected, in which nearly all frames has the same issue: they and their previous 'sweeps' have exactly the same timestamp.
 
https://github.com/open-mmlab/mmdetection3d/blob/1b39a483263825ac5a3e57fdf07d380adb7aa16d/tools/data_converter/waymo_converter.py#L236-L247

Tai-Wang:
Thanks for your suggestion. Yes, I agree with you. Because we actually do not play much with consecutive frames or use this timestamp information, we are not aware of this potential information loss. If you have some experience of it, could you please create a PR to fix it?

zhanggefan:
Hi,

Thank you for your response! 

I am now working on it and the PR is on the way. But I have a question I need to confirm with you in advance.

Could I change the output point cloud NumPy array's shape? 

If I could, I would remove the timestamp column, and store the mappings from the output bin file name to its timestamp in a separate file. 

- Reason: extending an output column for an identical value (waymo_converter.py L239) is obviously not efficient. It is even more problematic when generating waymo_infos_*.pkl, where for each bin file we need to read other 10 bin files to extract **ONLY** the timestamp. 

https://github.com/open-mmlab/mmdetection3d/blob/1b39a483263825ac5a3e57fdf07d380adb7aa16d/tools/data_converter/kitti_data_utils.py#L318-L319

https://github.com/open-mmlab/mmdetection3d/blob/1b39a483263825ac5a3e57fdf07d380adb7aa16d/tools/data_converter/kitti_data_utils.py#L418-L419

- Drawback: this would be a relatively big change because the code of WaymoDataset and config files would need to adapt as well. And tons of *.bin files would need to be regenerated again (harmful to the community cause many people spend days to convert those data).

Tai-Wang:
[...]

PR #1327, opened 2022-03-20
[Enhance] Parallel waymo dataset converter and ground truth database generator & Fix unprecise timestamps saving

zhanggefan:
This PR fixes issue #296. I substitute the timestamp column in the point cloud bin file to the element offset of each point from the original depth map. with this information included, we could easily rebuild the range view image in the data pipeline. This brings possibilities to re-implement range-view-based detectors in the future.

Meanwhile, this PR introduces a parallel waymo info gatherer and a parallel ground-truth database generator. The duration of the conversion of waymo dataset can be greatly shortened by paralleling processing.

**This PR does introduce BC-BREAK!!!!!** Because I modified the ground-truth database point cloud's format. 
**Now the newly generated database point clouds have 6 features per point rather than 5.**
The last feature is the element offset of each point from the original depth map. I packed this information into the database for the gt-augmentation of range-view-based methods.


tools/data_converter/kitti_data_utils.py
```diff
@@ -283,19 +294,9 @@ def map_func(idx):
...
-                         velodyne=False,
-                         calib=False,
-                         pose=False,
-                         image_ids=7481,
-                         extend_matrix=True,
-                         num_worker=8,
-                         relative_path=True,
-                         with_imageshape=True,
-                         max_sweeps=5):
+class WaymoImageInfoGatherer:
```
Tai-Wang, line 297: `WaymoInfoGatherer`

tools/data_converter/kitti_converter.py
```diff
@@ -187,55 +260,55 @@ def create_waymo_info_file(data_path,
...
         training=False,
         label_info=False,
         velodyne=True,
         calib=True,
         pose=True,
-        image_ids=test_img_ids,
         relative_path=relative_path,
-        max_sweeps=max_sweeps)
+        max_sweeps=max_sweeps,
+        num_worker=workers).gather(test_img_ids)
```
Tai-Wang, line 311: Split the class init and usage into two lines would be better?

tools/data_converter/kitti_converter.py
```diff
@@ -44,6 +44,78 @@ def _read_imageset_file(path):
     return [int(line) for line in lines]
 
 
+class _NumPointsInGTCalculater:
+    """Calculate the number of points inside the ground truth box. This is the
+    parallel version.
+
+    Args:
+        data_path (str): Path of the data.
+        relative_path (bool, optional): Whether to use relative path.
```
Tai-Wang, line 53: not optional?

Tai-Wang: or should update the following argument list?

tools/data_converter/kitti_converter.py
```diff
@@ -44,6 +44,78 @@ def _read_imageset_file(path):
...
+        Trv2c = calib['Tr_velo_to_cam']
+        P2 = calib['P2']
+        if self.remove_outside:
+            points_v = box_np_ops.remove_outside_points(
+                points_v, rect, Trv2c, P2, image_info['image_shape'])
+
+        # points_v = points_v[points_v[:, 0] > 0]
+        annos = info['annos']
+        num_obj = len([n for n in annos['name'] if n != 'DontCare'])
+        # annos = kitti.filter_kitti_anno(annos, ['DontCare'])
```
Tai-Wang, line 96: Can remove unused codes (and the above one)

tools/data_converter/kitti_converter.py
```diff
@@ -44,6 +44,78 @@ def _read_imageset_file(path):
     return [int(line) for line in lines]
 
 
+class _NumPointsInGTCalculater:
+    """Calculate the number of points inside the ground truth box. This is the
+    parallel version.
```
Tai-Wang, line 49: Can add a reference to the original version _calculate_num_points_in_gt.

Tai-Wang (commented):
Great job! I think it's ok to substitute the original unprecise timestamps with the mask indices, although my previous idea is to replace it with the frame_id because I remember the community has successful experiences using that for temporal modeling. Do you think which one is better? If we save mask indices in other ways, is there any better choice?

zhanggefan:
Hi, thanks for your suggestions!

> I think it's ok to substitute the original unprecise timestamps with the mask indices, although my previous idea is to replace it with the frame_id because I remember the community has successful experiences using that for temporal modeling. Do you think which one is better?

As for the choice between frame_id and timestamp for temporal modeling, I think both will work without much performance gap. But I think we'd better not place either timestamp or frame_id in the last column of the bin file. These values are invariant for each frame and duplicating these constants in a column is purely waste of storage. So I did these modifications to save a single copy of timestamp per frame into separate .txt files (in the timestamp subfolder) to save spaces. And these timestamps are read back by the info gathering code and then finally go into the pkl files. What's more, doing so will also boost the info gathering speed, because we do not need to read the full bin file back to get the timestamp. Instead, we only need to read the .txt file, which is much smaller in size.

>If we save mask indices in other ways, is there any better choice?

The mask indices are crucial to the range-view-based methods. And the indices are variant per point, so they must go into the bin files, otherwise, we would lose the grip of the correspondence between the indices and the points. 

There is another way. If we do not filter out any invalid points during the data conversion and do not concatenate the point cloud from the 5 lidar sources and instead store them into 5 separate files, then the range image can also be rebuilt by a reshape as long as we know the H and W of the range image. But this would introduce a huge bc-break and we would have to do a lot of refactors to the data pipeline to get it to work.

Tai-Wang (commented):
Almost ready. Due to bc-breaking changes, we need to add a corresponding instruction in the [compatibility doc](https://github.com/open-mmlab/mmdetection3d/blob/master/docs/en/compatibility.md). We would also add related content in the changelog of this version. (This PR only needs to modify the compatibility doc.)

Tai-Wang (approved):

ZwwWayne (approved):

PR #1435, opened 2022-04-27
Fix `NaiveSyncBatchNorm1d` and `NaiveSyncBatchNorm2d`

jshilong:
## Motivation
The `NaiveSyncBatchNorm1d` and `NaiveSyncBatchNorm2d` in the mmdet3d has two potential bugs.
- Could not do the training and inference when dist is unavailable
- `NaiveSyncBatchNorm1d` does not support the input with shape (N, C), which is not consistent with `nn.BatchNorm1d`

## Modification
- Add `using_dist` flag 
- Add flag `is_two_dim` to support the shape  (N, C),

## BC-breaking (Optional)
None


Tai-Wang (approved):

mmdet3d/ops/norm.py
```diff
@@ -53,11 +53,27 @@ def __init__(self, *args, **kwargs):
...
             f'input should be in float32 type, got {input.dtype}'
-        if dist.get_world_size() == 1 or not self.training:
+        using_dist = dist.is_available() and dist.is_initialized()
+        if (not using_dist) or dist.get_world_size() == 1 \
+                or not self.training:
             return super().forward(input)
         assert input.shape[0] > 0, 'SyncBN does not support empty inputs'
+        is_two_dim = input.dim() == 2
+        if is_two_dim:
+            input.unsqueeze_(2)
```
ZwwWayne, line 75: if we want to use inplace operation here, we should also use it in line 97.

ZwwWayne, line 75: if we want to use inplace operation here, we should also use it in line 97.

PR #1469, opened 2022-05-09
[Enhance] Support with_velocity=False in CenterPoint and NuScenesDataset

JoeyforJoy:
## Motivation
Fix some bugs of `CenterPoint` and `NuScenesDataset` when the velocity head in the `CenterHead`'s `common_heads` is ignored.

## Modification
- Simply add some conditional statements to support CenterPoint without the `vel` common head.
- Add a config file to test the no-velocity version of CenterPoint


mmdet3d/models/dense_heads/centerpoint_head.py
```diff
@@ -327,6 +327,8 @@ def __init__(self,
                 in_channels=share_conv_channel, heads=heads, num_cls=num_cls)
             self.task_heads.append(builder.build_head(separate_head))
 
+        self.with_velocity = 'vel' in common_heads.keys()
```
Tai-Wang, line 330: This can be written as a property of the class CenterPoint

mmdet3d/models/dense_heads/centerpoint_head.py
```diff
@@ -548,19 +554,31 @@ def get_targets_single(self, gt_bboxes_3d, gt_labels_3d):
...
-                        center - torch.tensor([x, y], device=device),
-                        z.unsqueeze(0), box_dim,
-                        torch.sin(rot).unsqueeze(0),
-                        torch.cos(rot).unsqueeze(0),
-                        vx.unsqueeze(0),
-                        vy.unsqueeze(0)
-                    ])
+                    if self.with_velocity:
+                        vx, vy = task_boxes[idx][k][7:]
+                        rot = task_boxes[idx][k][6]
```
Tai-Wang, line 559: line 559 to 562 can be moved out of the if-else

configs/centerpoint/centerpoint_02pillar_second_secfpn_4x8_cyclic_20e_nus_novelo.py
```diff
@@ -0,0 +1,185 @@
...
+        file_client_args=file_client_args),
+    dict(
+        type='LoadPointsFromMultiSweeps',
+        sweeps_num=9,
+        use_dim=[0, 1, 2, 3, 4],
+        file_client_args=file_client_args,
+        pad_empty_sweeps=True,
+        remove_close=True),
+    dict(type='LoadAnnotations3D', with_bbox_3d=True, with_label_3d=True),
+    # dict(type='ObjectSample', db_sampler=db_sampler),
```
Tai-Wang, line 85: Why do we remove this sampler? If necessary, can directly remove related lines.

mmdet3d/models/dense_heads/centerpoint_head.py
JoeyforJoy: If so, some checking codes would be better to be be added to avoid conflicts in the config file.

mmdet3d/models/dense_heads/centerpoint_head.py
JoeyforJoy: that's right

configs/centerpoint/centerpoint_02pillar_second_secfpn_4x8_cyclic_20e_nus_novelo.py
JoeyforJoy: I tried to use the sampler at first. If so, errors occur as bellow.
```
File "/home/ubuntu/repo/mmlab/PR/mmdetection3d/mmdet3d/datasets/pipelines/dbsampler.py", line 322, in sample_class_v2
    boxes = np.concatenate([gt_bboxes, sp_boxes], axis=0).copy()
  File "<__array_function__ internals>", line 5, in concatenate
ValueError: all the input array dimensions for the concatenation axis must match exactly, but along dimension 1, the array at index 0 has size 7 and the array at index 1 has size 9
```
This error may caused by the misalignment of the size between gt_bboxes and sp_boxes. 
It should be solved in future. But I'm not familiar with the sampler now, so I just commented the config file.

mmdet3d/models/dense_heads/centerpoint_head.py
Tai-Wang: I mean you can implement this line in a way like [this](https://github.com/open-mmlab/mmdetection3d/blob/master/mmdet3d/models/detectors/mvx_two_stage.py#L105)? We can keep the current logic to assign its value.

configs/centerpoint/centerpoint_02pillar_second_secfpn_4x8_cyclic_20e_nus_novelo.py
Tai-Wang: You can set the config of db_sampler to load the correct dimensions of gt_boxes.

JoeyforJoy: I add `bbox_code_size` attribute for `DataBaseSampler` to solve this problem in the new commit.

mmdet3d/models/dense_heads/centerpoint_head.py
JoeyforJoy: Good suggestion. I added it in the new commit.

mmdet3d/datasets/pipelines/loading.py
```diff
@@ -197,7 +199,8 @@ def __call__(self, results):
                     cloud arrays.
         """
         points = results['points']
-        points.tensor[:, 4] = 0
+        if points.tensor.shape[-1] >= 5:
+            points.tensor[:, 4] = 0
```
Tai-Wang, line 203: This hack is error-prone, please modify or add new options in the init to achieve this. Please also note to fix the unit test.

Tai-Wang (approved):

ZwwWayne (approved):

PR #1669, opened 2022-07-27
fix img shape

jshilong:
Fix the unitest of vote_fusion


ZCMax:
This ut bug has been fixed and merged into test-1.x

PR #1670, opened 2022-07-27
[Fix]Add seg mapping

jshilong:
Add `seg_label_mapping` for `Seg3DDataset` and `Det3DDataset` to map the label of the sematic mask to the valid class label.

### DIFF
--- a/mmdet3d/datasets/scannet_dataset.py
+++ b/mmdet3d/datasets/scannet_dataset.py
@@ -50,7 +50,12 @@ class ScanNetDataset(Det3DDataset):
         'CLASSES':
         ('cabinet', 'bed', 'chair', 'sofa', 'table', 'door', 'window',
          'bookshelf', 'picture', 'counter', 'desk', 'curtain', 'refrigerator',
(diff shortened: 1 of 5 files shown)


mmdet3d/datasets/scannet_dataset.py
```diff
@@ -67,6 +72,17 @@ def __init__(self,
...
+
+        # construct seg_label_mapping for semantic mask
+        seg_max_cat_id = self.METAINFO['seg_max_cat_id']
+        seg_valid_cat_ids = self.METAINFO['seg_valid_cat_ids']
+        neg_label = len(seg_valid_cat_ids)
+        seg_label_mapping = np.ones(
+            seg_max_cat_id + 1, dtype=np.int) * neg_label
+        for cls_idx, cat_id in enumerate(seg_valid_cat_ids):
+            seg_label_mapping[cat_id] = cls_idx
+        self.seg_label_mapping = seg_label_mapping
```
ZCMax, line 84: The `label_mapping` obtained in `Seg3DDataset` is actually a dict, not array. We need to unify them

PR #1682, opened 2022-08-01
[Fix] Remove MinkowskiEngine ImportError warnings

filaPro:
## Motivation

MinkowskiEngine is an extra requirement. We do not want to spam not using MinkowskiEngine users with warnings from all python files where it is imported (on every run of mmdetection3d).

## Modification

Just remove warning when MinkowskiEngine is not installed.

## BC-breaking (Optional)

No.

### DIFF
--- a/mmdet3d/datasets/pipelines/test_time_aug.py
+++ b/mmdet3d/datasets/pipelines/test_time_aug.py
@@ -138,6 +138,7 @@ class MultiScaleFlipAug3D(object):
             augmentation to point cloud. Defaults to True.
             Note that it works only when 'flip' is turned on.
     """
+
     def __init__(self,
                  transforms,
                  img_scale,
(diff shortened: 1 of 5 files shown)


ZwwWayne:
Lint issues should be fixed.

filaPro:
Can fix, but affected files are not modified in this PR...

Tai-Wang:
Yes, it's really strange, but could you please fix the problem mentioned in the lint check by the way? And If more problems appear afterward we will just directly merge this PR first.

filaPro:
I've somehow fixed lint, but it can be too annoying int the future... Should we synchronize lint configs in CI and `pre-commit-config.yml`?

Tai-Wang:
> I've somehow fixed lint, but it can be too annoying int the future... Should we synchronize lint configs in CI and `pre-commit-config.yml`? `mmcv` version in CI should probably also been fixed.

OK, I agree. We will fix them in the release PR.

Issue #1721, opened 2022-08-11
Behavior for no predictions during visualization


holtvogt:
**Describe the bug**
A `RuntimeError` is thrown during visualization with no predictions.

**Discussion**
I'd like to know how this error should be handled. As a check for `None` predicted labels is already done, I assume that the behavior on zero predictions should be the same. See below in **Bug fix**.

**Reproduction**
Use a model which does not predict labels on a test run.

1. What command or script did you run?

```
python ./tools/test.py <config.py> <checkpoint.pth> --show --show-dir <show-dir>
```

2. Did you make any modifications on the code or config? Did you understand what you have modified?

- Yes, I used my own config
- Yes, I understood

3. What dataset did you use?

- I used my own custom dataset

**Environment**

1. Please run `python mmdet3d/utils/collect_env.py` to collect necessary environment information and paste it here.

```
sys.platform: linux
Python: 3.8.13 (default, Mar 28 2022, 11:38:47) [GCC 7.5.0]
CUDA available: True
GPU 0: NVIDIA GeForce GTX 1080 Ti
CUDA_HOME: /usr/local/cuda
NVCC: Cuda compilation tools, release 11.6, V11.6.124
GCC: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
PyTorch: 1.10.1
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2021.4-Product Build 20210904 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.2.3 (Git Hash 7336ca9f055cf1bfa13efb658fe15dc9b41f0740)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.3
  - NVCC architecture flags: -gencode;arch=compute_37,code=sm_37;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_60,code=sm_60;-gencode;arch=compute_61,code=sm_61;-gencode;arch=compute_70,code=sm_70;-gencode;arch=compute_75,code=sm_75;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_37,code=compute_37
  - CuDNN 8.2
[...]

PR #1724, opened 2022-08-11
Fix visualization for empty prediction labels

holtvogt:
## Motivation

See #1721

## Modification

Added check for empty `pred_labels` tensor.


holtvogt:
Wrong initial branch.

PR #1854, opened 2022-09-20
[Feature] GT Augmentation Fade Strategy

AndyYuan96:

## Motivation

gtaug fade training strategy is very useful, which can get large improvement in many paper, but the strategy haven't implemented in mmdetection3d, for example , training total epoch is 20, last 5 epoch don't use gtaug, we need first training for 15 epochs, and then ctrl-c the training process, and the use resume_from to training for other 5epoch without gtaug, the process is not convenient. So I change the code, so that we don't need ctrl-c and resume training.

## Modification

I add a hook (EpochHook) to get current training epoch number, and use the number to update ObjectSample's cur_epoch , and I add a key (last_epoch) for ObjectSample, and I judge whether cur_epoch is larger than last_epoch in call function of ObjectionSample.

## BC-breaking (Optional)

## Use cases (Optional)

first add custom_hooks = [dict(type='EpochHook')] to config file, and then set dict(type='ObjectSample', db_sampler=db_sampler, last_epoch=15) in config file, last_epoch=15 means we use gtaug util epoch 15, and epoch 16 will not use gtaug.

## Checklist

2. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.3. 
4. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects.
5. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/configs/centerpoint/centerpoint_02pillar_second_secfpn_4x8_cyclic_20e_nus gtaugfade.py
+++ b/configs/centerpoint/centerpoint_02pillar_second_secfpn_4x8_cyclic_20e_nus gtaugfade.py
@@ -1,0 +1,172 @@
+_base_ = [
+    '../_base_/datasets/nus-3d.py',
+    '../_base_/models/centerpoint_02pillar_second_secfpn_nus.py',
+    '../_base_/schedules/cyclic_20e.py', '../_base_/default_runtime.py'
+]
+
+# If point cloud range is changed, the models should also change their point
+# cloud range accordingly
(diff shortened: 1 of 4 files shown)


Tai-Wang:
Hi @AndyYuan96 , thanks for your contribution! Please fix the lint failure (by using pre-commit hook following this [guide](https://github.com/open-mmlab/mmcv/blob/master/CONTRIBUTING.md)) and rebase your branch onto the `dev` instead of `master`. We typically merge new features into dev and dev-1.x for the stable and refactored version and then merge them into master and 1.x for the official release.

PR #1882, opened 2022-09-29
[Fix]: Fix nuscenes dataset training

Xiangxu-0103:
For 3D detection task on nuscenes dataset, the key `num_pts_feats` is not included in the `info`.

### DIFF
--- a/mmdet3d/datasets/det3d_dataset.py
+++ b/mmdet3d/datasets/det3d_dataset.py
@@ -26,11 +26,11 @@ class Det3DDataset(BaseDataset):
         metainfo (dict, optional): Meta information for dataset, such as class
             information. Defaults to None.
         data_prefix (dict, optional): Prefix for training data. Defaults to
-            dict(pts='velodyne', img="").
+            dict(pts='velodyne', img='').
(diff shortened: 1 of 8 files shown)


mmdet3d/datasets/det3d_dataset.py
```diff
@@ -246,7 +246,8 @@ def parse_data_info(self, info: dict) -> dict:
                     self.data_prefix.get('pts', ''),
                     info['lidar_points']['lidar_path'])
 
-            info['num_pts_feats'] = info['lidar_points']['num_pts_feats']
+            if 'num_pts_feats' in info['lidar_points']:
```
VVsssssk, line 249: Hi, I think num_pts_feats is an useful key for visualize, so can you add default num_pts_feats in convert script?
```python
temp_data_info['lidar_points']['num_pts_feats'] = 5
```
https://github.com/open-mmlab/mmdetection3d/blob/ea131ebc77bd91ecf91be8669c8702293b2390a3/tools/dataset_converters/update_infos_to_v2.py#L287

Tai-Wang (approved):

PR #2012, opened 2022-11-10
[CI] Add github action CI && refactor `tests/utils/`

JingweiZhang12:

## Motivation

## Modification

1. Add github action CI
2. Move `tests/utils/*.py` from `tests` to `mmdet3d/testing` for right importing on Win, like MMDet.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
4. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects.
5. The documentation has been modified accordingly, like docstring or example tutorials.


ZCMax:
We need to resolve the CUDA out of memory problem in GitHub action CI?

JingweiZhang12:
> We need to resolve the CUDA out of memory problem in GitHub action CI?

I tried to reduce memory footprint several mouths ago. Now, the occupancy is very small, but the RAM in the new GitHub CI is too small. The fix is a little challenging since we don't have Win env.

Tai-Wang (approved):

ZCMax (approved):

PR #2110, opened 2022-12-07
[Fix] lidar2cam error in update_infos_to_v2.py for nus and lyft dataset.

lianqing11:

## Motivation

lidar2cam error in update_infos_to_v2.py for nus and lyft dataset.

In the latest info version, we save the transformation from lidar to camera coordinate system with a unified transformation matrix, but the original implementation does not multiply the rotation matrix with the translation matrix, leading to errors in transformation matrices. This PR fixes this problem, and users should re-convert info with the new script.

This problem does not affect the currently supported models in 1.x but may yield potential errors for BEV-series detectors.

## Modification

change the translation from -t to -R^{-1}t

### DIFF
--- a/tools/dataset_converters/update_infos_to_v2.py
+++ b/tools/dataset_converters/update_infos_to_v2.py
@@ -302,8 +302,10 @@ def update_nuscenes_infos(pkl_path, out_dir):
                 ori_sweep['ego2global_rotation'],
                 ori_sweep['ego2global_translation'])
             lidar2sensor = np.eye(4)
-            lidar2sensor[:3, :3] = ori_sweep['sensor2lidar_rotation'].T
-            lidar2sensor[:3, 3] = -ori_sweep['sensor2lidar_translation']
+            rot = ori_sweep['sensor2lidar_rotation']
+            trans = ori_sweep['sensor2lidar_translation']
+            lidar2sensor[:3, :3] = rot.T
+            lidar2sensor[:3, 3] = -1 * np.matmul(rot.T, trans.reshape(3, 1))
             temp_lidar_sweep['lidar_points'][
                 'lidar2sensor'] = lidar2sensor.astype(np.float32).tolist()
             temp_lidar_sweep['timestamp'] = ori_sweep['timestamp'] / 1e6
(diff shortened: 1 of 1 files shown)


VVsssssk:
Hi,thanks your PR. I think you need to modify `lidar2sensor` in `lidar_sweeps` too. https://github.com/open-mmlab/mmdetection3d/blob/449ecb0e5ea65184f1b3e6fbd41e9a19478afa45/tools/dataset_converters/update_infos_to_v2.py#L306

VVsssssk:
And `lidar2sensor` in `image` dict. https://github.com/open-mmlab/mmdetection3d/blob/449ecb0e5ea65184f1b3e6fbd41e9a19478afa45/tools/dataset_converters/update_infos_to_v2.py#L333

lianqing11:
I have fixed the error in sweep.

Tai-Wang (approved):

Shiming94:
Hi, @lianqing11, @Tai-Wang , @ZwwWayne @VVsssssk, I still need to understand this change fully. 

```
## The code snippet from the nuscenes converter

    R = (l2e_r_s_mat.T @ e2g_r_s_mat.T) @ (
        np.linalg.inv(e2g_r_mat).T @ np.linalg.inv(l2e_r_mat).T)
    T = (l2e_t_s @ e2g_r_s_mat.T + e2g_t_s) @ (
        np.linalg.inv(e2g_r_mat).T @ np.linalg.inv(l2e_r_mat).T)
    T -= e2g_t @ (np.linalg.inv(e2g_r_mat).T @ np.linalg.inv(l2e_r_mat).T
                  ) + l2e_t @ np.linalg.inv(l2e_r_mat).T
```
I think `points_new = points @ R + T # here, R is already transposed` makes sense, as it can be calculated from:
`R_e2g @ (R_l2e @ points_new + T_l2e) + T_e2g = R_e2g_s @ (R_l2e_s @ points + T_l2e_s) + T_e2g_s`.
so why the `np.matul(R, T)` is needed here? 

Thank you a lot in advance. This question bothers me for a long time.

Issue #2112, opened 2022-12-07
[Bug] no vis_task parameter passed to add_datasample in hook


shanmo:
### Prerequisite

- [X] I have searched [Issues](https://github.com/open-mmlab/mmdetection3d/issues) and [Discussions](https://github.com/open-mmlab/mmdetection3d/discussions) but cannot get the expected help.
- [X] I have read the [FAQ documentation](https://github.com/open-mmlab/mmdetection3d/blob/1.1/docs/en/notes/faq.md) but cannot get the expected help.
- [X] The bug has not been fixed in the [latest version (dev)](https://github.com/open-mmlab/mmdetection3d/tree/dev) or [latest version (1.x)](https://github.com/open-mmlab/mmdetection3d/tree/dev-1.x).

### Task

I'm using the official example scripts/configs for the officially supported tasks/models/datasets.

### Branch

1.1x branch https://github.com/open-mmlab/mmdetection3d/tree/1.1

### Environment

sys.platform: linux
Python: 3.8.15 | packaged by conda-forge | (default, Nov 22 2022, 08:49:35) [GCC 10.4.0]
CUDA available: True
numpy_random_seed: 2147483648
GPU 0: NVIDIA T1200 Laptop GPU
CUDA_HOME: /home/mambaforge/envs/openmmlab
NVCC: Cuda compilation tools, release 11.7, V11.7.99
GCC: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
PyTorch: 1.13.0
PyTorch compiling details: PyTorch built with:
  - GCC 9.3
  - C++ Version: 201402
  - Intel(R) oneAPI Math Kernel Library Version 2022.1-Product Build 20220311 for Intel(R) 64 architecture applications
  - Intel(R) MKL-DNN v2.6.0 (Git Hash 52b5f107dd9cf10910aaa19cb47f3abf9b349815)
  - OpenMP 201511 (a.k.a. OpenMP 4.5)
  - LAPACK is enabled (usually provided by MKL)
  - NNPACK is enabled
  - CPU capability usage: AVX2
  - CUDA Runtime 11.7
[...]

PR #2118, opened 2022-12-08
[Fix] Fix missing parameter in `Det3DVisualizationHook`

shanmo:

## Motivation

Fix https://github.com/open-mmlab/mmdetection3d/issues/2112

## Modification

Add visualization task parameter in the hook

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/tools/test.py
+++ b/tools/test.py
@@ -29,6 +29,14 @@ def parse_args():
         'If specified, it will be automatically saved '
         'to the work_dir/timestamp/show_dir')
     parser.add_argument(
+        '--task',
+        type=str,
+        choices=[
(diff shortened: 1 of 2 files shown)


tools/test.py
```diff
@@ -28,6 +28,14 @@ def parse_args():
         help='directory where painted images will be saved. '
         'If specified, it will be automatically saved '
         'to the work_dir/timestamp/show_dir')
+    parser.add_argument(
```
ZwwWayne, line 31: why do need to specify task? If we use test.py, the visualized should be correctly initialized through configs and we should set it through config.

ZwwWayne: Thanks for your patient answer. We will merge this PR first, then consider making it simpler.

PR #2134, opened 2022-12-12
[Fix] Fix memory overflow in the rotated box IoU calculation

JingweiZhang12:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet3d/structures/bbox_3d/base_box3d.py
+++ b/mmdet3d/structures/bbox_3d/base_box3d.py
@@ -463,11 +463,17 @@ class BaseInstance3DBoxes(object):
         # height overlap
         overlaps_h = cls.height_overlaps(boxes1, boxes2)
 
+        # Restrict the min values of W and H to avoid memory overflow in
+        # ``box_iou_rotated``.
+        boxes1_bev, boxes2_bev = boxes1.bev, boxes2.bev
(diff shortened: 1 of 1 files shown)


VVsssssk (approved):
LGTM

mmdet3d/structures/bbox_3d/base_box3d.py
```diff
@@ -463,11 +463,17 @@ def overlaps(cls, boxes1, boxes2, mode='iou'):
         # height overlap
         overlaps_h = cls.height_overlaps(boxes1, boxes2)
 
+        # Restrict the min values of W and H to avoid memory overflow in
+        # ``box_iou_rotated``.
+        boxes1_bev, boxes2_bev = boxes1.bev, boxes2.bev
+        boxes1_bev[:, 2:4] = boxes1_bev[:, 2:4].clamp_min(1e-4)
```
ZwwWayne, line 469: clamp_min -> clamp(min=1e-4) is more common. Not sure whether clamp_min has full support of different pytorch versions.

filaPro:
Btw, @JingweiZhang12 
Do you remember the reason why this commit should fix the memory issue? And why it was (almost?) ok before this commit?

Actually last commit to this line before your commit was mine:) We are probably facing memory issue here now, but not sure. May be will check in the future.

Thanks in advance.

PR #2178, opened 2022-12-28
[Fix] Use int instead of np.int & Fix compability about spconv

JingweiZhang12:

## Modification

1. Use int instead of np.int 
2. Fix compability about spconv. Our compability about spconv has some problems:  
`hasattr(spconv, 'pytorch')` is always Flase if not implementing 'from spconv import pytorch'. 
So just remove this condition.
3. Not commenting `dbsample` in waymo dataset config.
4. Fix the datasetmeta loading bug.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist

3. The modification is covered by complete unit tests. If not, please add more unit test to ensure the correctness.
5. If the modification has potential influence on downstream projects, this PR should be tested with downstream projects.
6. The documentation has been modified accordingly, like docstring or example tutorials.

### DIFF
--- a/configs/_base_/datasets/waymoD5-3d-3class.py
+++ b/configs/_base_/datasets/waymoD5-3d-3class.py
@@ -38,7 +38,7 @@ db_sampler = dict(
 train_pipeline = [
     dict(type='LoadPointsFromFile', coord_type='LIDAR', load_dim=6, use_dim=5),
     dict(type='LoadAnnotations3D', with_bbox_3d=True, with_label_3d=True),
-    # dict(type='ObjectSample', db_sampler=db_sampler),
+    dict(type='ObjectSample', db_sampler=db_sampler),
     dict(
         type='RandomFlip3D',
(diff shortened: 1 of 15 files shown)


Xiangxu-0103:
@JingweiZhang12 We may also need to replace `np.bool` with `bool` to be compatible with the latest `numpy` version.

configs/_base_/datasets/waymoD5-3d-3class.py
```diff
@@ -38,7 +38,7 @@
 train_pipeline = [
     dict(type='LoadPointsFromFile', coord_type='LIDAR', load_dim=6, use_dim=5),
     dict(type='LoadAnnotations3D', with_bbox_3d=True, with_label_3d=True),
-    # dict(type='ObjectSample', db_sampler=db_sampler),
```
ZwwWayne, line 41: Why do we modify this line? We should do it in another PR.

Issue #1503, opened 2022-05-21
Fcos3D visualization error in the latest version


Etah0409:
I try to visualize Fcos3D (mini dataset). After I git clone the latest version, and fix some bugs.

/mmdet3d/datasets/nuscenes+mono_dataset.py
def show():
...
 img, img_metas = self._extract_data(i, pipeline,
                                                ['img', 'img_metas'])
            # need to transpose channel to first dim
            img = img.numpy().transpose(1, 2, 0)
...
the origin version will return NoneType to img_metas, and img is (900, 1600, 3) will be trans to (1600, 3, 900), which would cause a wrong mmcv.imwrite.

After I changed these 2 bugs to get the right img and img_metas, the visualization is like following:
![fe0b688a75cc6bca192260de4372d94](https://user-images.githubusercontent.com/29793682/169651857-510fde96-f574-47e7-97f4-231859e34e6a.png)

![e2a4807e2e1b8fbe591e7f1b371e337](https://user-images.githubusercontent.com/29793682/169651860-21829be3-7a2e-4936-9a99-2f6fe9d11eeb.png)

It seems that the size of bbox is wrong.
blue one is the pred-imgs, and red one is the gt. This result tests on Nuscenes mini dataset with fcos3d_finetune_model.
my cmd is :
python ./tools/test.py ./configs/fcos3d/fcos3d_r101_caffe_fpn_gn-head_dcn_2x8_1x_nus-mono3d.py ./work_dirs/fcos3d_r101_caffe_fpn_gn-head_dcn_2x8_1x_nus-mono3d_finetune_20210427_091419-35aaaad0.pth --eval mAP --eval-options show=True out_dir=./work_dirs/fcos3d_vis_mini

**Environment**
sys.platform: linux
Python: 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 06:04:18) [GCC 10.3.0]
CUDA available: True
GPU 0,1,2,3,4,5,6,7: NVIDIA RTX A6000
CUDA_HOME: /usr/local/cuda-11.1/
NVCC: Cuda compilation tools, release 11.1, V11.1.74
GCC: gcc (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
PyTorch: 1.8.1
PyTorch compiling details: PyTorch built with:
  - GCC 7.3
  - C++ Version: 201402
[...]

PR #2186, opened 2022-12-30
[CodeCamp #1503] Use mmeval.InstanceSeg

Pzzzzz5142:
## Motivation

Use mmeval.InstanceSeg

## Modification

+ mmdet3d/evaluation/functional/instance_seg_eval.py (deleted)
+ mmdet3d/evaluation/metrics/instance_seg_metric.py
+ tests/test_evaluation/test_metrics/test_instance_seg_metric.py

## BC-breaking

Delete the original implementation of instance_seg_eval.

## Use cases

For mmdetection3d currently doesn't support InstanceSeg, I cannot provide the use case now.


mmdet3d/evaluation/metrics/instance_seg_metric.py
```diff
@@ -1,33 +1,38 @@
...
-            If prefix is not provided in the argument, self.default_prefix
-            will be used instead. Default: None
+        dataset_meta (dict): Provide dataset meta information.
     """
 
     def __init__(self,
-                 collect_device: str = 'cpu',
+                 dataset_meta: dict,
+                 dist_backend: Optional[str] = None,
+                 collect_device: Optional[str] = None,
```
JingweiZhang12, line 23: Please add typehint

JingweiZhang12: My bad. I wan to say docstring about the input arguments.

JingweiZhang12 (commented):
Please make sure the UT of this part passes in your local env.

JingweiZhang12:
Please pull the latest code which has fix some errors about lint and UT.

filaPro:
(Sorry for stupid questions, but) Why `InstanceSegMetric` in `mmdetection3d` and `mmeval` should be different? And is unit test correct? (I mean single predicted instance mask should be split like [here](https://github.com/open-mmlab/mmeval/pull/70/files#diff-8bf38dad4097163ae6f7907600e17499952f493c2f69baf972feea46fdf8ddc8R33).)

Pzzzzz5142:
> (Sorry for stupid questions, but) Why `InstanceSegMetric` in `mmdetection3d` and `mmeval` should be different? And is unit test correct? (I mean single predicted instance mask should be split like [here](https://github.com/open-mmlab/mmeval/pull/70/files#diff-8bf38dad4097163ae6f7907600e17499952f493c2f69baf972feea46fdf8ddc8R33).)

This pr is try to use mmeval's metric. And it seems that we need to do some post-process on mmdetect3d's output in order to use mmeval's metric. I guess that if we use the same metric in `mmdetection3d` and `mmeval`, once `mmdetection3d` changes it output format, we also have to change the metric in `mmeval` which it is not a good idea for there may be other projects using the metric in `mmeval`. 

And the unit test is not correct currently in `mmdetection3d`, for we haven't determined the actual model output. However, I have verified that the metric should compute the right result.

sunjiahao1999:
@Pzzzzz5142 fix unit tests bug.

Pzzzzz5142:
Sorry for the delay (too busy recently), now I can pass the UT in my local env. 
image

PR #2223, opened 2023-01-15
[Feature] Add panoptic segmentation loading.

xizaoqu:

## Motivation

Add panoptic loading in LoadAnnotations3D for semanticKITTI and nuScenes. 

## Modification

Add _load_panoptic_seg_3d in LoadAnnotations3D. It simply loads panoptic labels from panoptic paths. The labels of segmentation datasets like semanticKITTI and nuScenes need further processing to get semantic labels and instance labels.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


mmdet3d/datasets/transforms/loading.py
```diff
@@ -718,6 +722,8 @@ class LoadAnnotations3D(LoadAnnotations):
             Defaults to False.
         with_seg_3d (bool): Whether to load 3D semantic masks for points.
             Defaults to False.
+        with_panoptic_seg_3d (bool): Whether to load 3D panoptic masks
```
ZwwWayne, line 725: with_panoptic_seg_3d -> with_panoptic_3d, for consistency with with_mask_3d and with_seg_3d

Xiangxu-0103:
It seems that we need to load `panoptic_annotation` with `seg_3d` and `mask_3d`. We should follow the implementation in [MMDetection](https://github.com/open-mmlab/mmdetection/blob/dev-3.x/mmdet/datasets/transforms/loading.py#L412),

xizaoqu:
> It seems that we need to load `panoptic_annotation` with `seg_3d` and `mask_3d`. We should follow the implementation in [MMDetection](https://github.com/open-mmlab/mmdetection/blob/dev-3.x/mmdet/datasets/transforms/loading.py#L412).

In datasets like semanticKITTI and nuScenes, both semantic labels and instance labels are in the same file, so I want to load them together in one function. For example, labels of semanticKITTI are constructed in the form of `(instance ids<<16)+semantic labels`.
And there are some questions. When only loading semantic labels, we need to filter out the instance part and do some mapping. When loading panoptic labels, we need to split the labels into semantic and instance parts. I plan to do it in the `LoadAnnotations3D` function with a new arg `offset`. What do you think of it?

Xiangxu-0103:
> > It seems that we need to load `panoptic_annotation` with `seg_3d` and `mask_3d`. We should follow the implementation in [MMDetection](https://github.com/open-mmlab/mmdetection/blob/dev-3.x/mmdet/datasets/transforms/loading.py#L412).
> 
> In datasets like semanticKITTI and nuScenes, both semantic labels and instance labels are in the same file, so I want to load them together in one function. For example, labels of semanticKITTI are constructed in the form of `(instance ids<<16)+semantic labels`. And there are some questions. When only loading semantic labels, we need to filter out the instance part and do some mapping. When loading panoptic labels, we need to split the labels into semantic and instance parts. I plan to do it in the `LoadAnnotations3D` function with a new arg `offset`. What do you think of it?

From my perspective, we need to split the original annotation label into `semantic_mask` and `instance_mask` as other datasets do for more common usage such as `semantic segmentation task`. Of course, more suitable method is also welcomed.

xizaoqu:
> > > It seems that we need to load `panoptic_annotation` with `seg_3d` and `mask_3d`. We should follow the implementation in [MMDetection](https://github.com/open-mmlab/mmdetection/blob/dev-3.x/mmdet/datasets/transforms/loading.py#L412).
> > 
> > 
> > In datasets like semanticKITTI and nuScenes, both semantic labels and instance labels are in the same file, so I want to load them together in one function. For example, labels of semanticKITTI are constructed in the form of `(instance ids<<16)+semantic labels`. And there are some questions. When only loading semantic labels, we need to filter out the instance part and do some mapping. When loading panoptic labels, we need to split the labels into semantic and instance parts. I plan to do it in the `LoadAnnotations3D` function with a new arg `offset`. What do you think of it?
> 
> From my perspective, we need to split the original annotation label into `semantic_mask` and `instance_mask` as other datasets do for more common usage such as `semantic segmentation task`. Of course, more suitable method is also welcomed.

Do you mean split labels with some preprocess? I think maybe it is not a good idea because it will occupy extra memory and therefore not friendly for user. Maybe I can implement one version and see if it suits.

Tai-Wang:
Please merge the latest dev-1.x and resolve the conflicts.

sunjiahao1999 (approved):
lgtm

mmdet3d/datasets/seg3d_dataset.py
```diff
@@ -107,14 +107,15 @@ def __init__(self,
...
         # construct seg_label_mapping for semantic mask
-        seg_max_cat_id = len(self.METAINFO['seg_all_class_ids'])
-        seg_valid_cat_ids = self.METAINFO['seg_valid_class_ids']
-        neg_label = len(seg_valid_cat_ids)
-        seg_label_mapping = np.ones(
-            seg_max_cat_id + 1, dtype=np.int) * neg_label
-        for cls_idx, cat_id in enumerate(seg_valid_cat_ids):
-            seg_label_mapping[cat_id] = cls_idx
-        self.seg_label_mapping = seg_label_mapping
+        if not hasattr(self, 'seg_label_mapping'):
```
ZwwWayne, line 110: We should implement a new function named `build_seg_label_mapping`, and call it in this place.
So that semantickitti_dataset could override this function and we could skip the check here

xizaoqu: I've fixed it.

PR #2236, opened 2023-01-28
[Feature] Support BEVFusion in `projects/`

JingweiZhang12:
## Motivation

We support the inference of BEVFusion in `projects/` 

## Modification

The BEVFusion mainly contains the backbones (image and lidar), LSS neck, and transformer-based head. 
Also, we add two ops: `bev_pool` and `voxelization`. **Note** that the voxelization OP in the original implementation of `BEVFusion` is different from the implementation in MMCV. If you want to use the original pretrained model [here](https://github.com/mit-han-lab/bevfusion/blob/main/README.md), you need to use the original implementation of voxelization OP.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


JingweiZhang12:
The pre-commit-config has been updated in this PR.

projects/BEVFusion/README.md
```diff
@@ -0,0 +1,143 @@
...
+<div align=center>
+<img src="https://user-images.githubusercontent.com/34888372/215313913-4b43f8a1-e2e2-49ba-b631-992155351922.png" width="800"/>
+</div>
+
+## Introduction
+
+We implement BEVFusion and provide the result and checkpoints on Waymo dataset.
+
+We follow the below style to name config files. Contributors are advised to follow the same style.
+`{xxx}` is required field and `[yyy]` is optional.
```
ZwwWayne, line 21: readers may not be able to get what is xxx and yyy, is this description missing something?
It should first tell users the config name style is `{model}_{model setting}_{backbone}_{neck}xxxx`

projects/BEVFusion/bevfusion/bevfusion.py
```diff
@@ -0,0 +1,328 @@
...
+        """
+        losses = self(**data)
+        loss, log_vars = self._parse_losses(losses)
+
+        outputs = dict(
+            loss=loss, log_vars=log_vars, num_samples=len(data['metas']))
+
+        return outputs
+
+    def val_step(self, data, optimizer):
```
ZwwWayne, line 142: make sure whether the modification of val_step and train_step is necessary

PR #2241, opened 2023-01-29
[Fix] Adjust the minimum supported python version to 3.7

Xiangxu-0103:

## Motivation

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


setup.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
```
ZwwWayne, line 1: this line seems to be an unnecessary modification

setup.py
```diff
@@ -23,13 +25,7 @@ def readme():
 def get_version():
     with open(version_file, 'r') as f:
         exec(compile(f.read(), version_file, 'exec'))
-    import sys
-
-    # return short version for sdist
-    if 'sdist' in sys.argv or 'bdist_wheel' in sys.argv:
```
ZwwWayne, line 29: these code might be necessary, we need to check whether the delete is safe.

JingweiZhang12:
Please pull the latest code, which has fixed the UT.

setup.py
Xiangxu-0103: I have checked that the delete is safe.

PR #2297, opened 2023-02-19
[Fix] Modify seg_3d_dtype from dtype to str

xizaoqu:

## Motivation

Can't set seg_3d_dtype to type such as 'np.int32' in config.

## Modification

Modify seg_3d_dtype from dtype to str.

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


ZwwWayne:
This is because np.int32 seems to be deprecated. Will it have any effect if it can not be np.int32?

xizaoqu:
> This is because np.int32 seems to be deprecated. Will it have any effect if it can not be np.int32?

It's just an example. It seems we can not set 'x = np.xxx' in the config.

PR #2340, opened 2023-03-09
[Fix] Fix numpy random state when import trimesh

sunjiahao1999:

## Motivation

I found the `trimesh` we used will change numpy random state.
```
>>> import numpy as np
>>> np.random.get_state()[2]
623
>>> import trimesh
>>> trimesh.__version__
'2.35.39'
>>> np.random.get_state()[2]
383
>>> 
```
But new version trimesh will not.
```
>>> import numpy as np
>>> np.random.get_state()[2]
623
>>> import trimesh
>>> trimesh.__version__
'3.20.2'
>>> np.random.get_state()[2]
623
```
It's usually not a problem when we use `register_all_modules()` because trimesh will import before `set_randomness()` in Runner. However, we remove `register_all_modules()` and use auto import in this [PR](https://github.com/open-mmlab/mmdetection3d/pull/2314) tirmesh will import when built `Det3DVisualizer` in hooks after `set_randomness()`. So we should fix this long-term potential bug.

## Modification

## BC-breaking (Optional)

## Use cases (Optional)

## Checklist


JingweiZhang12 (approved):

JingweiZhang12:
LGTM. BTY, the only place using `trimesh` is `write_oriented_bbox`, but this function is not called anywhere in our codebase currently.

PR #2509, opened 2023-05-08
[Update Documents] update dataset prepare documents about KITTI.

Raiden-cn:
## Motivation

When I used the KITTI dataset, I found that part of the description of the plane information in the data augmentation was missing.

## Modification

I have updated the related document in en and zh_cn.


JingweiZhang12:
@Raiden-cn Hi, this is a legacy issue. Good job! We'll appreciate it if you can regulate your PR following this [contributing guide](https://github.com/open-mmlab/mmdetection3d/blob/dev-1.x/docs/en/notes/contribution_guides.md).

PR #2527, opened 2023-05-14
[Fix] np.float as it is deprecated

filaPro:
np.float is removed from numpy

### DIFF
--- a/mmdet3d/evaluation/functional/scannet_utils/evaluate_semantic_instance.py
+++ b/mmdet3d/evaluation/functional/scannet_utils/evaluate_semantic_instance.py
@@ -25,8 +25,7 @@ def evaluate_matches(matches, class_labels, options):
     dist_confs = [options['distance_confs'][0]]
 
     # results: class x overlap
-    ap = np.zeros((len(dist_threshes), len(class_labels), len(overlaps)),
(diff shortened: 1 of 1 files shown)


mmdet3d/evaluation/functional/scannet_utils/evaluate_semantic_instance.py
```diff
@@ -25,8 +25,7 @@ def evaluate_matches(matches, class_labels, options):
     dist_confs = [options['distance_confs'][0]]
 
     # results: class x overlap
-    ap = np.zeros((len(dist_threshes), len(class_labels), len(overlaps)),
-                  np.float)
```
Xiangxu-0103, line 29: Suggest to change `np.float` to `np.float32`

filaPro: Looks like the default dtype for np.zeros is `float` which according to numpy documentation means exactly the same as old `np.float`. Why do we want to specify `np.float32`?

Xiangxu-0103: We expect to keep `dtype` to make the code more readable.

PR #2640, opened 2023-07-12
HardSimpleVFE.forward() doesn't need args coors 

WYYAHYT:
voxel coors are not involved with the process of HardSimpleVFE.forward(), so coors can be removed.

## Motivation
redudant args in code

## Modification
delete related args and the call to it.

### DIFF
--- a/mmdet3d/models/detectors/voxelnet.py
+++ b/mmdet3d/models/detectors/voxelnet.py
@@ -37,8 +37,7 @@ class VoxelNet(SingleStage3DDetector):
         """Extract features from points."""
         voxel_dict = batch_inputs_dict['voxels']
         voxel_features = self.voxel_encoder(voxel_dict['voxels'],
-                                            voxel_dict['num_points'],
-                                            voxel_dict['coors'])
+                                            voxel_dict['num_points'])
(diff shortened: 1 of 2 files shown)


mmdet3d/models/voxel_encoders/voxel_encoder.py
```diff
@@ -24,8 +24,8 @@ def __init__(self, num_features: int = 4) -> None:
         super(HardSimpleVFE, self).__init__()
         self.num_features = num_features
 
-    def forward(self, features: Tensor, num_points: Tensor, coors: Tensor,
```
sunjiahao1999, line 27: The change makes sense.

mmdet3d/models/detectors/voxelnet.py
```diff
@@ -37,8 +37,7 @@ def extract_feat(self, batch_inputs_dict: dict) -> Tuple[Tensor]:
         """Extract features from points."""
         voxel_dict = batch_inputs_dict['voxels']
         voxel_features = self.voxel_encoder(voxel_dict['voxels'],
-                                            voxel_dict['num_points'],
```
sunjiahao1999, line 40: You can't use HardVFE if you write it this way.

PR #2678, opened 2023-08-02
Update documentation in iou3d_calculator.py

ShaneXiangH:
## Motivation

The output `overlaps` shape in `mmdetection` has been [updated](https://github.com/open-mmlab/mmdetection/blob/f78af7785ada87f1ced75a2313746e4ba3149760/mmdet/structures/bbox/bbox_overlaps.py#L22). So the corresponding documentation should also be updated.

The logic of the code still work.

### DIFF
--- a/mmdet3d/structures/ops/iou3d_calculator.py
+++ b/mmdet3d/structures/ops/iou3d_calculator.py
@@ -41,8 +41,8 @@ class BboxOverlapsNearest3D(object):
 
         Return:
             torch.Tensor: If ``is_aligned`` is ``True``, return ious between
-                bboxes1 and bboxes2 with shape (M, N). If ``is_aligned`` is
-                ``False``, return shape is M.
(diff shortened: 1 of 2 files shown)


mmdet3d/structures/ops/iou3d_calculator.py
```diff
@@ -41,8 +41,8 @@ def __call__(self, bboxes1, bboxes2, mode='iou', is_aligned=False):
 
         Return:
             torch.Tensor: If ``is_aligned`` is ``True``, return ious between
-                bboxes1 and bboxes2 with shape (M, N). If ``is_aligned`` is
-                ``False``, return shape is M.
+                bboxes1 and bboxes2 with shape (M, N, 1). If ``is_aligned`` is
+                ``False``, return shape is (M, 1).
```
Xiangxu-0103, line 45: It seems that the comment is correct.

PR #2873, opened 2024-01-09
Centerpoint bbox coders: Fix torch warning

chbehrens:
## Motivation

I currently get the following warning from the centerpoints bbox coder:
```
mmdet3d/models/task_modules/coders/centerpoint_bbox_coders.py:207: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).
  self.post_center_range = torch.tensor(
```
The reason is that `self.post_center_range` is already a `torch.Tensor` after the first call of decode and then `torch.Tensor` is called on a Tensor in every following call. 

## Modification

Create `self.post_center_range` as Tensor in the initialization of `CenterPointBBoxCoder` and only make sure that it is on the right device when `decode` is called.

## BC-breaking (Optional)

No breaking/functionally relevant changes

## Use cases (Optional)

## Checklist

### DIFF
--- a/mmdet3d/models/task_modules/coders/centerpoint_bbox_coders.py
+++ b/mmdet3d/models/task_modules/coders/centerpoint_bbox_coders.py
@@ -36,7 +36,7 @@ class CenterPointBBoxCoder(BaseBBoxCoder):
         self.pc_range = pc_range
         self.out_size_factor = out_size_factor
         self.voxel_size = voxel_size
-        self.post_center_range = post_center_range
+        self.post_center_range = Tensor(post_center_range)
         self.max_num = max_num
         self.score_threshold = score_threshold
         self.code_size = code_size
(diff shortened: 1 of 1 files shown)


mmdet3d/models/task_modules/coders/centerpoint_bbox_coders.py
```diff
@@ -36,7 +36,7 @@ def __init__(self,
         self.pc_range = pc_range
         self.out_size_factor = out_size_factor
         self.voxel_size = voxel_size
-        self.post_center_range = post_center_range
+        self.post_center_range = Tensor(post_center_range)
```
Xiangxu-0103: Should notice that the default `post_center_range` is `None`, which may cause an error here.

chbehrens: True, would it be viable to throw the NotImplementedError already upon initialization or why is there a default value at all if it is currently required?

chbehrens: True, I updated it so that it stays None in that case.

chbehrens:
I added another small commit to fix the type annotation for `vel` in the decode function as it can be None as well as the dimensions of the `dim` and `vel` tensors in the docstring.

sunjiahao1999 (approved):

chbehrens:
Could someone please retrigger the pipelines that got stuck so this can get merged?

PR #2895, opened 2024-01-27
Update README.md

fmahdisoltani:
## Motivation

The command to run the `create_data.py` script, does not run in shell.

## Modification

Proper arguments should be passed to the script.

### DIFF
--- a/data/scannet/README.md
+++ b/data/scannet/README.md
@@ -11,7 +11,7 @@ We follow the procedure in [votenet](https://github.com/facebookresearch/votenet
 4. Enter the project root directory, generate training data by running
 
 ```bash
-python tools/create_data.py scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
+python tools/create_data.py --dataset scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
 ```
 
 The overall process could be achieved through the following script


data/scannet/README.md
```diff
@@ -11,7 +11,7 @@ We follow the procedure in [votenet](https://github.com/facebookresearch/votenet
 4. Enter the project root directory, generate training data by running
 
 ```bash
-python tools/create_data.py scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
+python tools/create_data.py --dataset scannet --root-path ./data/scannet --out-dir ./data/scannet --extra-tag scannet
```
Xiangxu-0103, line 14: It seems we do not need to add `--dataset` argument.

fmahdisoltani: I am not sure how you run this script, but on my end, when I run the script without the `--dataset` argument ( for scannet for example), I get the following error:


```
usage: create_data.py [-h] [--dataset kitti] [--root-path ROOT_PATH] [--version VERSION] [--max-sweeps MAX_SWEEPS] --out-dir OUT_DIR [--extra-tag EXTRA_TAG]
                      [--workers WORKERS]
create_data.py: error: unrecognized arguments: scannet
```

fmahdisoltani: I believe this [line](https://github.com/open-mmlab/mmdetection3d/blob/main/tools/create_data.py#L269) makes the script expecting the `--dataset` argument.

fmahdisoltani: I am using python 3.10 btw.

Issue #2510, opened 2023-05-09
Petr metricissues under version mmdetection3d-1.1.0


xiaowuge1201:
I trained PETR 24epoch through A800, but I cannot achieve the accuracy you announced.
The metric I received is：
mAP: 0.3131
mATE: 0.9197
mASE: 0.2800
mAOE: 0.6155
mAVE:0.8982
mAAE:0.2354
NDS: 0.3616
Can you publish the training log？？？

SekiroRong:
Here's my training log: [20221224_160645.log](https://github.com/open-mmlab/mmdetection3d/files/11496328/20221224_160645.log)

xiaowuge1201:
There is a problem with the Petr in the current mmdet3d-1.1.0 version, which may cause errors during runtime. Have you encountered this situation? You can use this version of code to run and test it; I made some minor corrections to the errors that occurred during the operation and conducted training, but the results were much worse than the ones you trained last year. Here is my training log：
[20230509_123941.log](https://github.com/open-mmlab/mmdetection3d/files/11506471/20230509_123941.log)，
Can you use this version for retraining at your convenience and submit the log for comparison? Thank you


> Here's my training log: [20221224_160645.log](https://github.com/open-mmlab/mmdetection3d/files/11496328/20221224_160645.log)

SekiroRong:
Could you please list the changes and errors? Unfortunately, I don't have the computation source to retrain PETR right now. However, you might be able to use the same version of this PETR implementation that I committed.

JingweiZhang12:
@xiaowuge1201 The links of the log and model have been uploaded in README.md in dev-1.x branch.

carlinds:
Hi! I also experience issues with reproducing the reported metrics. For reference, I get the following metrics when running a test on nuScenes validation split (with the config and model in the README):

```
mAP: 0.3829
mATE: 0.7376
mASE: 0.7096
mAOE: 0.4803
mAVE: 0.8703
mAAE: 0.2040
NDS: 0.3913
```

As you can see, the metrics differ from the reported ones, especially for mASE (and in turn NDS). 
The log you refer to seems to be from a much older state of the repo (it contains the now removed `LidarBox3dVersionTransfrom` for instance). It seems plausible that something has gone wrong when adapting the code for the coordinate system refactoring.

Did you run any model training and/or testing related to the following commits?
[...]

PR #2919, opened 2024-03-05
Fix PETR transforms and box handling to work with old checkpoints

carlinds:
## Motivation

Current version of the PETR-project does not produce the reported metrics (see https://github.com/open-mmlab/mmdetection3d/issues/2510 for more details). The current state of the project has the following issues:
- Incorrect rotation and scaling of the camera pose when doing 3D transforms.
- Incorrect normalizing (and denormalizing) of bounding boxes. Current normalizing is not compatible with the checkpoint in the repo, effectively switching length and width, causing large scale errors.

## Modification

This PR fixes:
- The 3D transforms to work correctly with versions >v1.0.0 (i.e after the coordinate system refactoring).
- The normalizing (and denormalizing) of boxes to work with the old checkpoint.

Results from testing the PR on nuScenes validation set (with the old checkpoint):
```
mAP: 0.3829
mATE: 0.7376
mASE: 0.2702
mAOE: 0.4803
mAVE: 0.8703
mAAE: 0.2040
NDS: 0.4352
```
Log from the test: [20240305_103838.log](https://github.com/open-mmlab/mmdetection3d/files/14494020/20240305_103838.log)

### DIFF
--- a/projects/PETR/petr/transforms_3d.py
+++ b/projects/PETR/petr/transforms_3d.py
@@ -175,35 +175,33 @@ class GlobalRotScaleTransImage(BaseTransform):
         return results
 
     def rotate_bev_along_z(self, results, angle):
-        rot_cos = torch.cos(torch.tensor(angle))
-        rot_sin = torch.sin(torch.tensor(angle))
+        rot_cos = np.cos(angle)
+        rot_sin = np.sin(angle)
 
-        rot_mat = torch.tensor([[rot_cos, -rot_sin, 0, 0],
-                                [rot_sin, rot_cos, 0, 0], [0, 0, 1, 0],
-                                [0, 0, 0, 1]])
-        rot_mat_inv = torch.inverse(rot_mat)
+        rot_mat = np.array([[rot_cos, rot_sin, 0, 0],
+                            [-rot_sin, rot_cos, 0, 0], [0, 0, 1, 0],
+                            [0, 0, 0, 1]])
+        rot_mat_inv = np.linalg.inverse(rot_mat)
         num_view = len(results['lidar2cam'])
         for view in range(num_view):
             results['lidar2cam'][view] = (
-                torch.tensor(np.array(results['lidar2cam'][view]).T).float()
-                @ rot_mat_inv).T.numpy()
+                results['lidar2cam'][view] @ rot_mat_inv)
 
         return
 
     def scale_xyz(self, results, scale_ratio):
-        rot_mat = torch.tensor([
+        scale_mat = np.array([
             [scale_ratio, 0, 0, 0],
             [0, scale_ratio, 0, 0],
             [0, 0, scale_ratio, 0],
             [0, 0, 0, 1],
         ])
 
-        rot_mat_inv = torch.inverse(rot_mat)
+        scale_mat_inv = np.linalg.inverse(scale_mat)
 
         num_view = len(results['lidar2cam'])
         for view in range(num_view):
-            results['lidar2cam'][view] = (torch.tensor(
-                rot_mat_inv.T
-                @ results['lidar2cam'][view].T).float()).T.numpy()
+            results['lidar2cam'][view] = (
+                scale_mat_inv @ results['lidar2cam'][view])
 
         return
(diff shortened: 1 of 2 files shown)


bradyz:
Wow great catch.
Have you tried training with these changes?

projects/PETR/petr/transforms_3d.py
```diff
@@ -175,35 +175,33 @@ def transform(self, results):
...
-        rot_mat = torch.tensor([
+        scale_mat = np.array([
             [scale_ratio, 0, 0, 0],
             [0, scale_ratio, 0, 0],
             [0, 0, scale_ratio, 0],
             [0, 0, 0, 1],
         ])
 
-        rot_mat_inv = torch.inverse(rot_mat)
+        scale_mat_inv = np.linalg.inverse(scale_mat)
```
TillBeemelmanns, line 200: `np.linalg.inverse()` does not exist `AttributeError: module 'numpy.linalg' has no attribute 'inverse'`
Should be `np.linalg.inv()`

projects/PETR/petr/transforms_3d.py
```diff
@@ -175,35 +175,33 @@ def transform(self, results):
...
+        rot_sin = np.sin(angle)
 
-        rot_mat = torch.tensor([[rot_cos, -rot_sin, 0, 0],
-                                [rot_sin, rot_cos, 0, 0], [0, 0, 1, 0],
-                                [0, 0, 0, 1]])
-        rot_mat_inv = torch.inverse(rot_mat)
+        rot_mat = np.array([[rot_cos, rot_sin, 0, 0],
+                            [-rot_sin, rot_cos, 0, 0], [0, 0, 1, 0],
+                            [0, 0, 0, 1]])
+        rot_mat_inv = np.linalg.inverse(rot_mat)
```
TillBeemelmanns, line 184: same here

TillBeemelmanns:
I trained PETR with the current code and with the suggestions in this PR. Both versions end up at the same metrics with only marginal differences:
```
NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4321
NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3766 
```
vs.
```
NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4353
NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3771
```

I would not merge this PR, but rather provide new weights for the current setup.

bradyz:
here's a little weight conversion script -
let me know if i missed something, i only did a quick qualitative check

```python
import torch

ckpt = torch.load('weights/petr.pth')

keys = [
    'pts_bbox_head.reg_branches.0.4.weight',
    'pts_bbox_head.reg_branches.1.4.weight',
    'pts_bbox_head.reg_branches.2.4.weight',
    'pts_bbox_head.reg_branches.3.4.weight',
    'pts_bbox_head.reg_branches.4.4.weight',
    'pts_bbox_head.reg_branches.5.4.weight',
]

for k in keys:
    orig = ckpt['state_dict'][k]

    # weights are (10, 256)
    modified = orig.clone()
    modified[2:3] = orig[3:4]
    modified[3:4] = orig[2:3]

    ckpt['state_dict'][k] = modified

torch.save(ckpt, 'weights/petr_fixed.pth')
```

carlinds:
> I trained PETR with the current code and with the suggestions in this PR. Both versions end up at the same metrics with only marginal differences:
> 
> ```
> NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4321
> NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3766 
> ```
> 
> vs.
> 
> ```
> NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4353
> NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3771
> ```
> 
> I would not merge this PR, but rather provide new weights for the current setup.



> I trained PETR with the current code and with the suggestions in this PR. Both versions end up at the same metrics with only marginal differences:
> 
> ```
> NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4321
> NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3766 
> ```
> 
> vs.
> 
> ```
> NuScenes metric/pred_instances_3d_NuScenes/NDS: 0.4353
> NuScenes metric/pred_instances_3d_NuScenes/mAP: 0.3771
> ```
> 
> I would not merge this PR, but rather provide new weights for the current setup.

Providing new weights would also suffice. However, I would still suggest keeping some of the fixes in this request (such as the incorrect rotation, misleading naming and unnecessary conversion between torch and numpy).

PR #3058, opened 2024-11-21
[Fix] Problem of failing to save images after open3d visualization

Aiuan:
## Motivation

While verifying the inference demo according to the following code in [Get Started](https://mmdetection3d.readthedocs.io/en/latest/get_started.html),  we often find that the open3d library can visualize the point cloud and detection results, but after closing the visualization window, the results are not saved to the image (the image is often completely black). To fix this issue we raised this PR.

Code:
```
python demo/pcd_demo.py demo/data/kitti/000008.bin pointpillars_hv_secfpn_8xb6-160e_kitti-3d-car.py hv_pointpillars_secfpn_6x8_160e_kitti-3d-car_20220331_134606-d42d15ed.pth --show
```

Visualize window:
image

Saved image (before modification):
image

## Modification

We modify the `show()` function in the mmdet3d/visualization/local_visualizer.py file.

1. We annotate `self.o3d_vis.clear_geometries()` in line 887, because geometries should not be cleared before saving as an image.
2. We modify `self.o3d_vis.capture_screen_image(save_path)` in line 896 to  `self.o3d_vis.capture_screen_image(save_path, do_render=True)`,  because we need to re-render after manually closing the visualization window to ensure that the stored image is not empty.

After our modification, the visualization results can be saved to the image normally.

Saved image (after modification):
image

### DIFF
--- a/mmdet3d/datasets/transforms/dbsampler.py
+++ b/mmdet3d/datasets/transforms/dbsampler.py
@@ -280,7 +280,7 @@ class DataBaseSampler(object):
                 s_points_list.append(s_points)
 
             gt_labels = np.array([self.cat2label[s['name']] for s in sampled],
-                                 dtype=np.long)
+                                 dtype=np.int64)
 
             if ground_plane is not None:
                 xyz = sampled_gt_bboxes[:, :3]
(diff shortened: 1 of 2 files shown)


Puiching-Memory:
Thank you. It worked. They haven't had any activity for almost a year. Do you know what happened?

Puiching-Memory:
https://github.com/open-mmlab/mmdetection3d/pull/3058/files#diff-caaf06b4c95f877cef520cf35f7f0ba25f750a862c7dca1933421d4c1efbc40bL894-R898

I recommend adding self.o3d_vis.clear_geometries() after the exported image, because the current modification causes two images to be rendered in a row, and the result of the previous image is saved to the next one

PR #3140, opened 2026-02-11
[Fix] Replace hardcoded .cuda() with device-aware .to() in MinkUNet voxelization

Mr-Neutr0n:
## Motivation

In `Det3DDataPreprocessor.voxelize()`, the `minkunet` voxelization branch converts numpy arrays back to tensors using `torch.from_numpy(...).cuda()`. This hardcodes the CUDA device assumption, which causes:

1. **RuntimeError on CPU-only environments** — users without a GPU cannot use MinkUNet-based models at all
2. **Incorrect device placement in multi-GPU setups** — tensors always land on `cuda:0` regardless of which device the input data resides on (e.g., `cuda:1`), leading to device mismatch errors during subsequent operations

## Modification

Replace `.cuda()` with `.to(res.device)` for both `point2voxel_map` and `inds` tensors, where `res` is the input point cloud tensor already on the correct device. This is consistent with how device handling is done elsewhere in the same file (e.g., using `.new_tensor()` and `F.pad()` which inherit the device from existing tensors).

**Before:**
```python
point2voxel_map = torch.from_numpy(point2voxel_map).cuda()
...
inds = torch.from_numpy(inds).cuda()
```

**After:**
```python
point2voxel_map = torch.from_numpy(point2voxel_map).to(res.device)
...
inds = torch.from_numpy(inds).to(res.device)
```

## BC-breaking (No)

This is a backward-compatible fix. On CUDA environments, `res.device` will be `cuda:X` (matching the previous behavior when on `cuda:0`), and it additionally supports CPU and multi-GPU scenarios correctly.

### DIFF
--- a/mmdet3d/models/data_preprocessors/data_preprocessor.py
+++ b/mmdet3d/models/data_preprocessors/data_preprocessor.py
@@ -434,12 +434,12 @@ class Det3DDataPreprocessor(DetDataPreprocessor):
                 res_coors_numpy = res_coors.cpu().numpy()
                 inds, point2voxel_map = self.sparse_quantize(
                     res_coors_numpy, return_index=True, return_inverse=True)
-                point2voxel_map = torch.from_numpy(point2voxel_map).cuda()
+                point2voxel_map = torch.from_numpy(point2voxel_map).to(res.device)
                 if self.training and self.max_voxels is not None:
                     if len(inds) > self.max_voxels:
                         inds = np.random.choice(
                             inds, self.max_voxels, replace=False)
-                inds = torch.from_numpy(inds).cuda()
+                inds = torch.from_numpy(inds).to(res.device)
                 if hasattr(data_sample.gt_pts_seg, 'pts_semantic_mask'):
                     data_sample.gt_pts_seg.voxel_semantic_mask \
                         = data_sample.gt_pts_seg.pts_semantic_mask[inds]
--- a/mmdet3d/models/data_preprocessors/data_preprocessor.py
+++ b/mmdet3d/models/data_preprocessors/data_preprocessor.py
@@ -434,7 +434,8 @@ class Det3DDataPreprocessor(DetDataPreprocessor):
                 res_coors_numpy = res_coors.cpu().numpy()
                 inds, point2voxel_map = self.sparse_quantize(
                     res_coors_numpy, return_index=True, return_inverse=True)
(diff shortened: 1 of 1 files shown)


Mr-Neutr0n:
I have read the CLA Document and I hereby sign the CLA

Mr-Neutr0n:
The 2-line fix in `Det3DDataPreprocessor.voxelize()` is correct and reads cleanly against the surrounding code. Verified the surrounding context: `res` is a point cloud tensor that has already been moved to the right device earlier in the function (it gets `.new_tensor(voxel_size)` etc. further down), so `res.device` is exactly what you want to round-trip back to.

The `pr_stage_test` failure pattern matches what we saw on #3869 — the same CircleCI matrix is failing on the most restrictive env (PyTorch 1.8.1 / Python 3.7). The `.cuda()` → `.to(res.device)` change is fully backward-compatible:
  - CUDA single-GPU: `res.device == cuda:0`, same as before
  - CUDA multi-GPU: now uses the right device (this is the actual fix)
  - CPU: now works (this is the actual fix)
  - ROCm/MPS: now uses the right device instead of failing on `.cuda()`

It cannot have *introduced* a new failure on the 1.8.1/3.7 matrix because the semantic at the call sites is unchanged for any single-GPU CUDA run. Almost certainly a pre-existing flake on that old combo or an unrelated dependency issue.

Worth re-running the `pr_stage_test` workflow (or pushing an empty commit to retrigger) — happy to push the empty commit if you give the go-ahead.

Mr-Neutr0n:
Pushed a no-op commit (`91a2fc6`) to retrigger `pr_stage_test` on the current branch. The actual code change is still the same 2-line `.cuda()` → `.to(res.device)` fix; this is just to get a fresh run on the flaky old-matrix job.

Mr-Neutr0n:
Pushed a fix for the failing CI. Wrapped the `point2voxel_map = torch.from_numpy(point2voxel_map).to(res.device)` assignment across two lines in `data_preprocessor.py` to fix flake8 E501 (82 > 79 characters) introduced by PR #3140’s `.cuda()` → `.to(res.device)` change, which was failing CircleCI lint and blocking `pr_stage_test`; readthedocs failures appear unrelated to this code change.

Mr-Neutr0n:
Hi @ZwwWayne - a gentle nudge on this device-aware fix for MinkUNet; still mergeable and applies cleanly. Happy to adjust anything, and fine to close if it is not wanted. Thank you.

